PR #45, opened 2019-05-28
Support different object types in registry service

### LINKED ISSUE #41: Extend Schema registry to support additional objects
- [ ] Support for additional objects in rest api (policies, endpoints, oidc providers, etc...), maybe use a more generic approach
- [ ] Update grpc to include all objects in response
- [ ] Validate endpoint should support validation of different objects

AleF83:
Closes #41


services/registry/src/sync/sync-service.ts
```diff
@@ -1,60 +1,63 @@
...
-        {}
-      )
-    )
-  ),
-  combineAll(),
-  map(schemas => schemas.reduce((acc, next) => ({ ...acc, ...next }))),
-  shareReplay(1)
+    map(([sourceName, source]) =>
+        // FIXME: Not only schemas
+        defer(() => source.getGqlObjects("gqlschemas")).pipe(
```
Yshayy: Let's change the getGqlObject to return all source data. (no "kind" parameter).
After we get the combined RawConfigurationData, we can apply several field transformers (schema, endpoints, etc...) to create the final GqlConfigurationMessage

PR #51, opened 2019-06-18
Make gql schema object to be extendable

AleF83:


services/registry/src/sources/index.ts
```diff
@@ -1,8 +1,9 @@
 import fetch from "node-fetch";
+import { GqlAgogosObjectConfig } from "../sync/object-types";
 
 export default interface Source {
-    getGqlObjects(): Promise<{ [kind: string]: { [name: string]: string } }>;
-    putGqlObject(name: string, kind: string, definition: string): Promise<void>;
+    getGqlObjects(): Promise<{ [kind: string]: { [name: string]: GqlAgogosObjectConfig } }>;
```
Yshayy: Change to AgogosObjectConfig. Change source interface correspondingly

PR #70, opened 2019-07-30
Improve resources validation

### LINKED ISSUE #60: Schema validation feature is broken
It's probably mismatching names, mostly (`gqlschemas` vs `schemas` etc)
Probably an artifact of #54

AleF83:
This PR fixes validation flow that is being invoked during creating new Agogos object.

Closes #60


services/registry/src/validation/index.ts
```diff
@@ -0,0 +1,24 @@
...
+    upstream: validateUpstream,
+    upstreamclientcredentials: validateUpstreamClientCredentials,
+};
+
+
+export const validateNewObject = async (name: string, kind: string, source: string, spec: AgogosObjectConfig): Promise<void> => {
+    if (!validators.hasOwnProperty(kind)) {
+        throw new Error("Unknown GraphQL object kind");
+    }
+    await validators[kind].call(null, source, name, spec);
```
AvivRubys: Why the `.call`?

services/registry/src/sync/sync-grpc-server.ts
```diff
@@ -1,96 +1,97 @@
...
+                // TODO: extract to function
+                const upstreams: Upstream[] = Object.values(
+                    configuration.upstreams
+                ).map(ep => {
+                    const upstream = new Upstream();
+                    upstream.setHost(ep.host);
 
-          const upstreamAuth = new UpstreamAuthentication();
-          // FIXME: get from ep
-          upstreamAuth.setAuthType(ep.auth.authType);
```
AvivRubys: Types say `auth`, CRD says `authType`
Verify this is correct please :)

services/registry/src/validation/validators/upstreamValidator.ts
```diff
@@ -0,0 +1,24 @@
...
+import { AgogosObjectConfig, UpstreamConfig } from "../../sync/object-types";
+
+const validateUpstream = async (
+    source: string,
+    name: string,
+    spec: AgogosObjectConfig
+): Promise<void> => {
+    const upstreams = await gqlObjects$
+        .pipe(
+            map(x => x.upstreams || {}),
```
AvivRubys: Does this `map`->`take(1)` pattern used in all validators not block deployment when agogos is up but doesn't have any CRDs deployed?

services/registry/src/sync/sync-grpc-server.ts
AleF83: It works so I think all things are correct

services/registry/src/validation/validators/upstreamValidator.ts
AleF83: There is `startWith` [here](https://github.com/Soluto/agogos/blob/7c60e86149d9e84920a6431a2b313f90dfaf9492/services/registry/src/sync/sync-service.ts#L70).
So it shouldn't be empty ever

AvivRubys (approved):

PR #73, opened 2019-07-31
TSLint: Configure and refactor code

AleF83:


AvivRubys:
The "enforcing all interfaces to begin with I" and "all types must be interfaces" rules are just changing half this codebase for no reason without any benefit =\

AvivRubys (approved):

PR #76, opened 2019-08-04
Reuse existing oauth tokens in graphql server

AvivRubys:


tomeresk (approved):
Are there no cases where the cached token expires or needs to be invalidated?

AleF83 (changes requested):
Great idea!
I just think maybe not to store all token sources in one cache but as a property of clientCredentials object.
So it can be more generic.

AvivRubys:
@tomeresk the `oauth2` library [takes care of that](https://github.com/golang/oauth2/blob/master/clientcredentials/clientcredentials.go#L80)
@AleF83 good suggestion, see latest commit. This also has the benefit of not needing to hash anything - we can just use the scope as the id, because the cache is bound to the rest of the parameters anyway since it's on the same object.

AleF83 (approved):

PR #81, opened 2019-08-13
Select and Alias directives

AvivRubys:
* Better URL construction (did not escape the strings)
* ~Added `@alias` directive~
* Added `@select` directive
* Added concurrent resolver support to the result transform middleware


Yshayy:
Aren't the two operations similar?
Maybe:
@select(field:String)
@selectPath(path:[String])

AvivRubys:
They definitely are, at first I mistakenly used just select before realising they are actually different
This is why I don't like making them so similar though - it can be confusing to use.
How about `@select(path: [String])` and `@from(field: String)`?

yoavkarako:
How about `@property`?

yoavkarako:
Just a thought, if you want to go out of your way for this directives.
You can merge them under a single name, then use either the `leaf` or the `result_transformer` depending on if there's already a leaf directive on the field.

AvivRubys:
I think having two nullable fields and making you choose between them seems just as confusing as two similar names, but with more places you can go wrong in, like passing in two nulls :X
`@property` sounds good but... Is kind of really abstract and doesn't really mean anything I guess?
I'm gonna rename to `@from` because it's already miles better than `@alias`

yoavkarako:
> I think having two nullable fields and making you choose between them seems just as confusing as two similar names

Use a single field. There's no reason both use-cases can't use with a composite path. It makes the `Lead` case more versatile.

AvivRubys:
1. Having one directive act in two different ways sounds like something that can easily become hard to understand
2. Using a single array field makes no sense - you'll never need to alias a deep field because that would mean it's inside a whole other type

AvivRubys:
I'm going to drop `@from` for now, since it has some buggy interactions with sibiling resolvers, which can make it very confusing to use.
For example, when `@from` and `@http` on two sibiling fields, the `@http` resolver can't use the target field name of the `@from`'d field, only the original field name.
This is a fundamental issue because resolvers can rely on the parent resolver, not on sibiling resolvers.

AleF83 (approved):

PR #83, opened 2019-08-18
Atomic registry updates

AvivRubys:
* No globals
* New ServerContext struct/interface holding the schema extension objects - just upstreams right now


AvivRubys:
I haven't changed what it does, just slightly how.
Anyway after this is merged I'll open a PR for the main package refactor (it's already done but i wanna rebase it on master), which will remove schema.go entirely anyway

PR #85, opened 2019-08-19
Improve registry schema validation

AvivRubys:
Now running validation beyond basic parsing on the schema, should catch things like misused/unknown directives, bad field/type definitions, etc


services/registry/src/sync/sync-service.ts
```diff
@@ -15,7 +15,7 @@ import sources from "../sources-config";
 import logger from "../logger";
 import { AgogosObjectConfig } from "./object-types";
 
-export type AggObjsByName = { [name: string]: AgogosObjectConfig }
+export type AggObjsByName<Obj extends AgogosObjectConfig = AgogosObjectConfig> = { [name: string]: AgogosObjectConfig }
```
AleF83:
```suggestion
export type AggObjsByName<Obj extends AgogosObjectConfig = AgogosObjectConfig> = { [name: string]: Obj }
```

services/registry/src/sync/sync-schemas.ts
```diff
@@ -44,9 +42,20 @@ export const makeGqlDocumentFromGqlSources = (gqlSchemas: AggObjsByName) => {
     return mergeAllDocuments(documentNodes);
 };
 
+const validateSchemasAsBool = (schemas: AggObjsByName<SchemaConfig>) => {
```
AleF83: I don't like this name. Maybe `isSchemaValid`

AleF83 (changes requested):
Consider cancelling the whole configuration update on invalid schema

tomeresk (approved):

PR #91, opened 2019-08-27
Add better service mocks

AleF83:
This PR adds some mock services for better future testing.
Also I changed tslint to eslint with tslint plugin in e2e and mock services (Inspired by @tomeresk )


e2e/mocks/customer-api/Dockerfile
```diff
@@ -0,0 +1,24 @@
+# Stage 1: dependencies
+FROM node:10-alpine as dependencies
+WORKDIR /app
+COPY package.json ./
```
tomeresk: Should also copy the lock file here and in build

e2e/mocks/order-api/Dockerfile
```diff
@@ -0,0 +1,24 @@
+# Stage 1: dependencies
+FROM node:10-alpine as dependencies
+WORKDIR /app
+COPY package.json ./
```
tomeresk: need lock file here as well

e2e/package.json
```diff
@@ -3,7 +3,7 @@
     "version": "1.0.0",
     "license": "MIT",
     "scripts": {
-        "test": "mocha --require ts-node/register ./**/*.spec.ts --exit",
+        "test": "mocha --require ./tsnodeRegister.js ./**/*.spec.ts --exit",
         "tslint-check": "tslint-config-prettier-check ./tslint.json"
```
tomeresk: Should probably also change this to eslint

e2e/mocks/order-api/.eslintrc
```diff
@@ -0,0 +1,31 @@
...
+        "space-before-function-paren": "off",
+        "space-in-parens": "off",
+        "array-bracket-spacing": "off",
+        "semi": "off",
+        "@typescript-eslint/semi": ["error"]
+    },
+    "overrides": [{
+        "files": ["test/**"],
+        "rules": {
+            "no-unused-expressions": "off"
```
tomeresk: This override is to allow chai expressions like `expect(something).to.exist` which the linter would normally disallow (which is why it is only applied on files under test directory).
Since chai is not used in `order-api`' we can optionally remove this

e2e/mocks/customer-api/.eslintrc
```diff
@@ -0,0 +1,31 @@
...
+        "space-before-function-paren": "off",
+        "space-in-parens": "off",
+        "array-bracket-spacing": "off",
+        "semi": "off",
+        "@typescript-eslint/semi": ["error"]
+    },
+    "overrides": [{
+        "files": ["test/**"],
+        "rules": {
+            "no-unused-expressions": "off"
```
tomeresk: This override is to allow chai expressions like `expect(something).to.exist` which the linter would normally disallow (which is why it is only applied on files under test directory).
Since chai is not used in `customer-api`' we can optionally remove this

e2e/.eslintrc
```diff
@@ -0,0 +1,31 @@
...
+    "rules": {
+        "@typescript-eslint/no-unused-vars": "error",
+        "space-before-function-paren": "off",
+        "space-in-parens": "off",
+        "array-bracket-spacing": "off",
+        "semi": "off",
+        "@typescript-eslint/semi": ["error"]
+    },
+    "overrides": [{
+        "files": ["test/**"],
```
tomeresk: This limits the override to only apply to files under the `test` folder, though the folder in this cases is `tests` so we should change the name here

tomeresk (approved):

PR #94, opened 2019-08-28
JSON scalar

AvivRubys:
* Added JSON scalar
* Removed `log` and `overrideContext` unused test directives


tomeresk (approved):

services/graphql-server/scalars/json_literal_test.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func TestObject(t *testing.T) {
+	val1 := ast.NewIntValue(nil)
+	val1.Value = "123"
+
+	val2 := ast.NewIntValue(nil)
+	val2.Value = "1337"
+
+	list := ast.NewListValue(nil)
```
AleF83: Should it be `NewObjectValue`?

AleF83 (changes requested):

services/graphql-server/scalars/json_literal_test.go
AvivRubys: It should. Don't copy and paste code, kids.

AleF83 (approved):

PR #97, opened 2019-08-30
Graphql directive

AleF83:
This PR adds new directive that allows to `gateway` to resolve data from remote graphql server like it already does from remote rest server with `@http` directive


services/graphql-server/directives/common/gql.go
```diff
@@ -0,0 +1,86 @@
...
+
+	upstream, ok := s.Upstream(url.Host)
+	if ok {
+		upstream.ApplyUpstream(rp.Context, &request.Header)
+	}
+	return request, nil
+}
+
+func sendRequest(request *gqlclient.Request, client *gqlclient.Client, gqlParams gqlParams) (interface{}, error) {
+	ctx := context.Background()
```
AvivRubys: Use context from ResolveParams

services/graphql-server/directives/common/gql.go
```diff
@@ -0,0 +1,86 @@
...
+	var respData interface{}
+	if err := client.Run(ctx, request, &respData); err != nil {
+		logrus.WithError(err).Panic("error while graphql request")
```
AvivRubys: No reason to panic when we can just return the error or wrap it

services/graphql-server/directives/common/gql.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	result := respData.(map[string]interface{})[gqlParams.queryName]
```
AvivRubys: Possible panic here, check type conversion with second argument

services/graphql-server/utils/ast2sdl.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/graphql-go/graphql"
+	"github.com/graphql-go/graphql/language/ast"
+)
+
+// ResolveParamsToSDLQuery - generates SDL query from graphql.ResolveParams object
+func ResolveParamsToSDLQuery(queryName string, rp graphql.ResolveParams) string {
```
AvivRubys: This could use a unit test or two

services/registry/Dockerfile
```diff
@@ -2,6 +2,11 @@ FROM node:10-alpine as base
 WORKDIR /service
 
 FROM base as dependencies
+
+RUN apk add --no-cache --virtual .build-deps make gcc g++ python \
```
AvivRubys: What is this for?

AleF83: node-gyp

services/registry/builtin-resources/schemas/base.gql
```diff
@@ -24,7 +24,7 @@ directive @select(path: [String!]!) on FIELD_DEFINITION
 directive @gql(
     url: String!
     queryName: String!
-    arguments: String
+    arguments: String = ""
```
AvivRubys: Why is this a string? Shouldn't it be some kind of key-value mapping?

e2e/tests/11_graphql_directive_tests.spec.ts
```diff
@@ -48,22 +49,60 @@ describe("Graphql Directive tests", () => {
...
+                        reviews: reviews.filter(r => r.hotelId === o.hotelId).map(r => ({
+                            author: r.author,
+                            text: r.text,
+                        })),
+                    },
+                })),
+            },
         };
 
         const response = await client.request(`{
```
AvivRubys: Use `graphql-tag` so you can use variables easily without escaping your own strings and such

services/registry/builtin-resources/schemas/base.gql
AleF83: I'll take care of this in next PR.

services/graphql-server/utils/ast2sdl.go
AleF83: I'll take care of this in next PR.

e2e/tests/11_graphql_directive_tests.spec.ts
AleF83: I'll take care of it in next PR. I want to add support for non-graphql resolvers under graphql directive anyway.

Issue #88, opened 2019-08-22
Schema testing


AvivRubys:
How does a team that wants to expose data on agogos, or upload a new version of a schema, test their changes?
Testing it in isolation is easy, but if it's referencing other schemas this gets dicey.
We trust the registry's validation to protect the schema from corruption, so maybe we could say that deploying IS testing.
But I'm not sure that's enough. Particularly if you're editing an existing schema, testing to prevent accidentally breaking your contract, for example, has a lot of value.

AvivRubys:
Idea courtesy of @Yshayy - make it easy to run agogos locally and load schemas into it. We can create a docker image with a simple filesystem-based resource system.

AvivRubys:
Mostly closed by #98

PR #98, opened 2019-09-04
Standalone Agogos

AvivRubys:
To support local testing (#88) for our consumers:
* Added kubernetes-fs source instead of the existing filesystem source, so we could add in proper production schemas and test them easily
* Added `soluto/agogos-standalone` image, which should let anyone run agogos locally with resources read from a local directory, in order to test/develop their schemas and other resources.


remote-sources/kubernetes-fs/main.go
```diff
@@ -0,0 +1,123 @@
...
+
+var unknownKind = "UNKNOWN_KIND"
+
+func toIndex(resource AgogosResource) string {
+	switch resource.Kind {
+	case "UpstreamClientCredentials":
+		return "upstreamclientcredentials"
+	case "Upstream":
+		return "upstreams"
+	case "Schema":
```
tomeresk: Might be worth putting these mappings in a map and returning `unknownKind` if missing in the map:

```go
var indices = map[string]string{
	"UpstreamClientCredentials": "upstreamclientcredentials",
	"Upstream": "upstreams",
	"Schema": "schemas",
}
    
func toIndex(resource AgogosResource) string {
	index, ok := indices[resource.Kind]
	if !ok {
		index = unknownKind
	}
	
	return index
}
```

remote-sources/kubernetes-fs/main.go
```diff
@@ -0,0 +1,123 @@
...
+	default:
+		return unknownKind
+	}
+}
+
+func toResourceFormat(resources []AgogosResource) map[string]map[string]AgogosResourceSpec {
+	result := make(map[string]map[string]AgogosResourceSpec)
+
+	for _, resource := range resources {
+		index := toIndex(resource)
```
tomeresk: Maybe `toIndex` should have a second `ok` return value instead of `unknownKind`? feels like the idiomatic go way for it

tomeresk (approved):

tomeresk (approved):

Yshayy:
kubernetes-fs was designed to run k8s resources without actual k8s?

AvivRubys:
Yes, to make it easy for our users to test their schemas locally

PR #103, opened 2019-09-12
Add fragments and variables support for @gql directive 

AleF83:


services/graphql-server/directives/common/gql.go
```diff
@@ -78,8 +78,11 @@ func createGqlClient(url string) *gqlclient.Client {
 }
 
 func createGqlRequest(s server.ServerContext, gqlParams gqlParams, rp graphql.ResolveParams) (*gqlclient.Request, error) {
-	query := utils.ResolveParamsToSDLQuery(gqlParams.queryName, rp, gqlParams.args)
-	request := gqlclient.NewRequest(query)
+	requestConfig := utils.ResolveParamsToSDLRequest(gqlParams.queryName, rp, gqlParams.args)
```
AvivRubys: It's odd that we pass in the "raw" args before the variable insertion. `ResolveParamsToSDLRequest` should have one job, which is taking the queryName&ast and returning the SDL (with extra stuff like variables fragments etc if needed)

services/graphql-server/utils/ast2sdl.go
```diff
@@ -2,72 +2,119 @@ package utils
...
-// ResolveParamsToSDLQuery - generates SDL query from graphql.ResolveParams object
-func ResolveParamsToSDLQuery(queryName string, rp graphql.ResolveParams, args string) string {
-	builder := strings.Builder{}
+// GqlRequestConfig - type to store all data for gql request
+type GqlRequestConfig struct {
+	Query          string
+	VariableValues map[string]interface{}
+}
+
+type sdlQuery struct {
```
AvivRubys: Unclear on two very similarly named fields - `fragmentsToAdd` and `addedFragments`

services/graphql-server/utils/ast2sdl.go
```diff
@@ -81,3 +128,50 @@ func buildCustomArgumentsClause(builder *strings.Builder, rp graphql.ResolvePara
...
+		query.addedFragments[fragmentName] = true
+		query.fragmentsToAdd = query.fragmentsToAdd[1:]
+		fragment := rp.Info.Fragments[fragmentName].(*ast.FragmentDefinition)
+
+		query.builder.WriteString(fmt.Sprintf("fragment %s on %s", fragmentName, fragment.TypeCondition.Name.Value))
+		resolveParamsToQueryInner(query, fragment, rp)
+	}
+}
+
+func buildVariableClause(query *sdlQuery) string {
```
AvivRubys: Everything here is called `buildXYZ` and it straight up adds stuff to the string builder
This one returns a string instead? And I'm not really sure what it does

services/graphql-server/utils/ast2sdl.go
```diff
@@ -2,72 +2,119 @@ package utils
...
+type GqlRequestConfig struct {
+	Query          string
+	VariableValues map[string]interface{}
+}
+
+type sdlQuery struct {
+	builder        *strings.Builder
+	fragmentsToAdd []string
+	addedFragments map[string]bool
+	variables      map[string]string
```
AvivRubys: I thought `variables` was key-value but it looks like it's key-type? I'm a bit confused

services/graphql-server/utils/ast2sdl.go
AleF83: Variable clause is located in the beginning of the query but can be build only after all query is built so I used another builder for it and then it the whole query string is built from variable clause and fields and other parts

services/graphql-server/utils/ast2sdl.go
AleF83: One fragment can be dependent on another one and several fragments can be dependent from one. So I need to manage what fragments I need to add to query and what fragment already was added. So `fragmentToAdd` stores fragments that should be added to query and `addedFragments` contains fragments that already were added to query string

services/graphql-server/utils/ast2sdl.go
AvivRubys: Sure that's fine, but all the other `buildXYZ` methods add stuff to the string builder on the `sdlQuery` struct, whereas this one returns a string.
If they're all called `buildXYZ` I'd expect them to behave the same way, and they don't.

services/graphql-server/utils/ast2sdl.go
AvivRubys: So we track all fragments we run into while traversing the AST, and when we finish we append the fragments clauses?
Sounds like we should only need one non-duplicate list of all fragments we run into for that, no? If we append them all at once at the end, why do we need to track which ones we've already added?

AleF83: I'll try to explain by example.
1) Query refers to fragments A and B
2) Fragment A refers no other fragments
3) Fragment B refers to fragment A and C.
4) Fragment C reference no fragments. 

After building query we have A and B in list.
Then we add A and B fragment clauses to the whole string.
During addition of B we see that there is reference to A and C.
But A fragment was already added to the result string and C wasn't
So in addition to the set (a.k.a. non-duplicate list) we need something that can tell us if the fragment was already added or need to be added.

services/graphql-server/utils/ast2sdl.go
AleF83: I can rename it to `getVariableClause`. Does it be ok for you?

services/graphql-server/utils/ast2sdl.go
AvivRubys: Talked offline, we settled on creating a struct that will abstract the map-array weirdness with some methods like "SeenFragment" "AddedFragment" etc

services/graphql-server/utils/ast2sdl.go
AvivRubys: Talked offline, decided on renaming the other functions to `writeXYZ` instead of `buildXYZ`

services/graphql-server/utils/ast2sdl.go
AvivRubys: Talked offline, decided on finding a better name for `variables` that expresses it's dual purpose as 1. variables that were referenced in the query 2. variable-to-type mapping

AleF83:
@AvivRubys can you review again please?

PR #112, opened 2019-10-22
rest directive

### LINKED ISSUE #95: Re-design @http directive
Current:
* Response assumed to always be json
* Serializes "input" arg as the body if it exists
* Different syntax for variable replacement in url and in query/body/headers
```graphql
### DEFINITION
directive @http(
    url: String!
    method: String
    contentType: String # body type. "json" or null
    query: [HttpNameValue!] # key value pairs with arg/source/etc insertion
    body: [HttpNameValue!] # When you want to construct a body with variables from args/source/etc? I think
    timeout: Int
    headers: [HttpNameValue!] # key value pairs with arg/source/etc insertion
) on FIELD_DEFINITION
```

Proposal:
* Introduce `StringMap` scalar that'll just be stringy key-value pairs, to make header/query params easier to use
* Change variable replacement syntax - with inspiration from [apollo](https://www.apollographql.com/docs/link/links/rest/#rest-directive), replace the `:id` syntax with `{source/args/context.id}`
  * No ambiguity about which variable will be inserted
  * Uniform syntax across url, query params and headers for both static and variable-inserted values
* Has feature parity, besides constructing the body from variables from context/source/etc.

```graphql
## DEFINITION
# response is deserialized by content-type
directive @http(
    url: String!
    method: String
    bodyArg: String # The name of the argument to take the body from. defaults to "input"
    bodySerializer: BodySerializers # Request body serialization method. defaults to Json
    headers: StringMap
    query: StringMap
    timeout: Int
) on FIELD_DEFINITION

enum BodySerializers {
    UrlEncoded
    Json
    Xml
}

scalar StringMap

## USAGE:
type Query {
  user(id: ID!, company: String): User @http(url: 'http://myurl/{args.id}`, query: {company: "{args.company}"}, headers: {"x-api-client": "agogos"})
}

type Mutation {
  newUser(input: NewUserInput): User @http(url: 'http://myurl/', bodySerializer: UrlEncoded)
}
```

tomeresk:
Closes #95


services/graphql-server/directives/common/rest/rest.go
```diff
@@ -0,0 +1,265 @@
...
+	"net/url"
+	"strings"
+	"time"
+
+	"github.com/graphql-go/graphql"
+	"github.com/sirupsen/logrus"
+	"github.com/vektah/gqlparser/ast"
+)
+
+type RestParams struct {
```
AvivRubys: Should this be exported?

services/graphql-server/directives/common/rest/rest.go
```diff
@@ -0,0 +1,265 @@
...
+
+	for _, arg := range fieldArgs {
+		nameValuePair := arg.(map[string]interface{})
+		nameValueList.append(nameValuePair["name"].(string), nameValuePair["value"].(string))
+	}
+
+	return nameValueList
+}
+
+func createHTTPClient(timeoutMs int) http.Client {
```
AvivRubys: I'm not sure, but I thought something like `7 * time.Millisecond` should just work, without needing to break it down to nanoseconds and back again

services/graphql-server/directives/common/rest/rest.go
tomeresk: Not really no, I'll change it

AvivRubys (commented):
I have a half-baked idea about NameValueList and the array resolving stuff, which is basically separating the replacement into two parts - parse template into "injectable" and static parts, and resolving values of a parsed template or parts of it.
That way we can use the entire thing when we just want to replace everything (like in url), and areas like query where arrays are important could use the parts separately and combine them a bit differently.
But this is half-baked as it is, if the current implementation works we can return to this as a refactor later on.

I think this can maybe also make sense in the GQL directive implementation, where we're gonna need to inject the variable with the correct type in, not always coerce it to a string.

services/graphql-server/directives/common/rest/rest.go
tomeresk: `Duration` expects its input to be in nanoseconds, so I'm converting it from ms to ns, but I'm not converting it back at any point

services/graphql-server/directives/common/rest/rest.go
```diff
@@ -0,0 +1,265 @@
...
+		return "", err
+	}
+
+	addQueryParamsToURL(urlObject, queryParams, resolveParams)
+
+	finalURL := urlObject.String()
+	return finalURL, nil
+}
+
+func addQueryParamsToURL(urlObject *url.URL, queryParams NameValueList, resolveParams graphql.ResolveParams) {
```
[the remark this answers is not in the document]
tomeresk: Fair enough, I changed it to return the querystring

services/graphql-server/utils/param_replacement.go
```diff
@@ -32,15 +35,40 @@ func resolveFromParams(rp graphql.ResolveParams, sourceName string, propName str
...
+	sepIndex := strings.IndexRune(template, '.')
+
+	sourceName := template[1:sepIndex]
+	propName := template[sepIndex+1 : len(template)-1]
+
+	return resolveFromParams(rp, sourceName, propName)
+}
+
+// GetReplacementIfArray - If str is only a single template that resolves to an array, it returns that array, otherwise it returns nil
+func GetReplacementIfArray(rp graphql.ResolveParams, str string) []interface{} {
```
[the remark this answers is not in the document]
tomeresk: I do want the name to also reflect that the array is not actually created, it only happens when the input replacement param itself is an array.
How about `ResolveSingleArrayTemplate`?

services/graphql-server/directives/common/rest/rest.go
AvivRubys: Yes, but `7 * time.Millisecond` is already a `time.Duration` by the magic of golang types

services/graphql-server/utils/param_replacement.go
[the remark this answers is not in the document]
AvivRubys: 👍  sounds good

services/graphql-server/directives/common/rest/rest.go
tomeresk: That magic only works with number literals, but as much as I could find it doesn't work when the number you need comes from a variable (`timeoutMs`)

AvivRubys: https://play.golang.org/p/yqxfp4znTd0
Seems to work?

AvivRubys (approved):

PR #113, opened 2019-10-30
"exportAs" directive

AvivRubys:
* Refactored middlewares to take a MiddlewareContext object instead of 4 arguments
* Added `exportAs` directive


tomeresk (changes requested):
We should probably also add at least a simple e2e test that uses it

tomeresk (dismissed):

tomeresk (approved):

PR #120, opened 2020-01-29
Adding auth

CHarnel:


src/modules/auth/index.ts
```diff
@@ -6,15 +6,21 @@ export async function getAuthHeaders(context: RequestContext, url: URL) {
...
+                    credentials.activeDirectory.clientId,
+                    credentials.activeDirectory.clientSecret,
+                    upstream.auth.activeDirectory.resource
+                );
+
+                return {
+                    Authorization: `Bearer ${token}`,
+                };
+            } catch (ex) {
+                console.error(ex);
```
AvivRubys: Use `./modules/logger` with some error message

src/modules/directives/gql.ts
```diff
@@ -1,20 +1,38 @@
...
 export class GqlDirective extends SchemaDirectiveVisitor {
     visitFieldDefinition(field: GraphQLField<any, any>) {
         const {url, fieldName, arguments: gqlArgs, operationType: operationTypeParam} = this.args;
         const operationType = operationTypeParam?.toLowerCase() ?? 'query';
-        const link = new HttpLink({uri: url, fetch: fetch as any});
-        const remoteSchema = introspectSchema(link).then(schema => makeRemoteExecutableSchema({schema, link}));
+
+        const http = new HttpLink({uri: url, fetch: fetch as any});
+        let schemaInitialized = false;
+        let remoteSchema = {} as GraphQLSchema;
```
AvivRubys: Can we do something like `let remoteSchema : GraphQLSchema | null` and avoid using the `schemaInitialized` variable with this?

src/modules/directives/gql.ts
```diff
@@ -1,20 +1,38 @@
...
         field.resolve = async (parent, args, context, info) => {
+            if (!schemaInitialized) {
+                const link = setContext(async () => ({
+                    headers: await getAuthHeaders(context, new URL(url)),
+                })).concat(http);
+
+                remoteSchema = await introspectSchema(link)
+                    .then(schema => makeRemoteExecutableSchema({schema, link}))
+                    .catch(ex => {
+                        console.error(ex);
```
AvivRubys: Use `./modules/logger` with some error code

PR #125, opened 2020-02-25
multiple resource repositories

CHarnel:


services/src/modules/resource-repository/local.ts
```diff
@@ -0,0 +1,27 @@
...
+    fetch(): Promise<ResourceGroup>;
+    async fetch(): Promise<ResourceGroup | null> {
+        const json = await fs.readFile(this.localPath, 'utf8');
+        return JSON.parse(json);
+    }
+
+    async update(rg: ResourceGroup): Promise<void> {
+        const {etag, ...rgWithoutEtag} = rg;
+        await fs.writeFile(this.localPath, JSON.stringify(rgWithoutEtag));
+    }
```
AvivRubys: encoding specified when reading but not when writing

services/src/modules/resource-repository/repositoryManager.ts
```diff
@@ -0,0 +1,38 @@
...
+    const repos = initRepos();
+    const resources = await Promise.all(repos.map(repo => repo.fetch(currentEtag)));
+    if (resources.length == 0) return null;
+    let newRg: ResourceGroup = {schemas: [], upstreams: [], upstreamClientCredentials: []};
+
+    for (const rg of resources) {
+        if (rg != null) {
+            newRg = applyResourceGroupUpdates(newRg, rg);
+        }
+    }
```
AvivRubys: This bit is almost exactly `reduce`

AvivRubys:
I'm doing a refactor which will make this process easier, but don't worry, the code will live on 😎

AvivRubys:
I'm doing a refactor which will make this process easier, but don't worry, the code will live on 😎

PR #133, opened 2020-04-30
Authorization - fully implemented registry part

tomeresk:
This includes:

- Create/update policy resource
- Attachments support for policy resource (with support for writing the attachment to both s3 and fs repositories)
- Opa policy type implementation, including compiling rego code to wasm and adding that to the policy as an attachment


services/src/modules/resource-repository/fs.ts
```diff
@@ -26,12 +28,30 @@ export class FileSystemResourceRepository implements ResourceRepository {
         await fs.writeFile(this.pathToFile, JSON.stringify(rg));
     }
 
+    async writePolicy(filename: string, compiledRego: Buffer): Promise<void> {
```
Yshayy, line 31: I think that compileRego adds a bit of dependency, maybe it should abstract as a regular resource and add attachment support? (maybe with path conventions, or getAttachment or something)

tomeresk: I agree about `compiledRego`, it was an oversight on my part, I will rename it to something more generic.

Regarding the policy, it is already a regular resource that is saved in the resources file as part of the resource group. This function only writes the attachment (compiled rego in the case of opa) and is generic, so it can work as attachment for other types as well (poor variable naming aside).
After explaining it now I realize the function name itself (`writePolicy`) is also bad since it doesn't write the policy itself (which is written with the regular `update` function) but only the attachment. I will rename the function itself as well.

Do you still think other changes are required here other than renaming the `compiledRego` variable to something like `content` and the `writePolicy` function to something like `writePolicyAttachment`?

As a note, the attachment itself is stored with a name convention based on the metadata of namespace and name, there is no explicit link to it in the policy resource

tomeresk: @Yshayy This is addressed in commit 738408c

AvivRubys (approved):

services/src/registry.ts
```diff
@@ -144,8 +222,36 @@ async function fetchAndValidate(updates: Partial<ResourceGroup>): Promise<Resour
     return newRg;
 }
 
+type tempLocalPolicyAttachment = {metadata: ResourceMetadataInput; path: string; type: PolicyType};
```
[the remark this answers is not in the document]
tomeresk: I can make the change, the standard for type naming is Pascal case?

PR #141, opened 2020-06-18
Add policy definitions and attachments to request context, change pol…

tomeresk:
…icy executor to use them from context instead of directly from repo

### DIFF
--- a/services/src/modules/resource-repository/types.ts
+++ b/services/src/modules/resource-repository/types.ts
@@ -3,6 +3,7 @@ export interface ResourceGroup {
     upstreams: Upstream[];
     upstreamClientCredentials: UpstreamClientCredentials[];
     policies: Policy[];
+    // policyAttachments are compiled from the Rego code in opa policies, they are not directly modified by users
     policyAttachments?: PolicyAttachments;
 }
 
--- a/services/src/modules/resource-repository/types.ts
+++ b/services/src/modules/resource-repository/types.ts
@@ -3,6 +3,7 @@ export interface ResourceGroup {
     upstreams: Upstream[];
     upstreamClientCredentials: UpstreamClientCredentials[];
     policies: Policy[];
+    policyAttachments?: PolicyAttachments;
 }
 
 export interface Resource {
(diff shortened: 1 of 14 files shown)


services/src/modules/resource-repository/types.ts
```diff
@@ -3,6 +3,7 @@ export interface ResourceGroup {
     upstreams: Upstream[];
     upstreamClientCredentials: UpstreamClientCredentials[];
     policies: Policy[];
+    policyAttachments?: PolicyAttachments;
```
AvivRubys, line 6: Leave a comment about this being a "runtime-only" property? Since the rest are assumed to always exist

services/src/tests/integration/registry/create-resources.spec.ts
```diff
@@ -67,7 +67,13 @@ const policy = {
     ],
 };
 
-const baseResourceGroup = {schemas: [], upstreams: [], upstreamClientCredentials: [], policies: []};
+const baseResourceGroup = {
+    schemas: [],
+    upstreams: [],
+    upstreamClientCredentials: [],
+    policies: [],
+    policyAttachments: {},
```
AvivRubys, line 75: Shouldn't `policyAttachments` not be here?

services/src/tests/integration/resourceBucket.ts
```diff
@@ -17,13 +18,29 @@ export function mockResourceBucket(initialValue: ResourceGroup, initialPolicyFil
         .reply(200, (_, body) => {
             value.current = JSON.parse(body as string) as ResourceGroup;
         })
+        .get(new RegExp(`/${bucketName!}\?.*prefix=${encodeURIComponent(policiesKeyPrefix!)}.*`))
```
AvivRubys, line 21: Perhaps a comment or some constants could make this more readable, because right now it's a bit hard :X

AvivRubys (commented):
LGTM Generally, commented on some small things

services/src/tests/integration/registry/create-resources.spec.ts
tomeresk: It is technically there because of the way we currently manage the bucket mocking, but I'll remove it to be in line with the tests in `update-resources.spec` (since this is not something we really test here)

services/src/modules/resource-repository/types.ts
tomeresk: Come to think of it, it will be populated on the first call to `fetchLatest` same as the others, and will always exist (empty object if there are no attachments at all, same as the other fields here).
All of them are actually `runtime-only` properties representing what is stored in the underlying repository, with the only different between the attachments and the other properties being that the other properties are all stored in one file and the attachments are stored in other files.

However, when using this interface we don't really care that it is stored in other files than the other properties, so I'm actually leaning towards removing the `?` here and treating it like any other property.

One key difference is that the contents of this property are calculated based on the `policies` property and not created/updated directly by a user, but it only happens once when we save the policy and then saved this way in the repository, so I would still say we should treat it like the other properties here (but I will add a comment to highlight this difference)

AleF83 (approved):

AvivRubys (approved):

PR #143, opened 2020-06-25
Support for policy query

AleF83:


services/src/tests/e2e/tests/authorization_with_queries.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+        });
+        expect(schemaResponse.updateSchemas.success).toBeTruthy();
+
+        // Wait for gateway to update before next tests
+        await sleep(500);
+    });
+
+    test('Query allowed employee', async () => {
+        const response: AllowedEmployeeQueryResponse = await gatewayClient.request(employeeQuery('allowedEmployee'));
+        expect(response.allowedEmployee.address).toBeDefined();
```
tomeresk, line 185: Why not check the actual value of the address? should be pretty simple here

AleF83: I'll add snapshot validation

services/src/modules/paramInjection.ts
```diff
@@ -9,27 +10,27 @@ type jwtData = {
...
+            return parent && R.path(propPath, parent);
         case 'args':
-            return args && args[propName];
+            return args && R.path(propPath, args);
         case 'exports':
-            return context.exports.resolve(info.parentType, parent, propName);
+            return context.exports.resolve(info.parentType, parent, key);
         case 'jwt':
-            return getJwt(context)[propName];
+            return getJwt(context)[key];
```
tomeresk, line 33:
```suggestion
            return R.path(propPath, getJwt(context));
```

We can support the deep nesting here as well.
`getJwt` will return an empty object if there is not JWT so there is no need for an initial existence check.

PR #147, opened 2020-07-01
Eslint

AleF83:


Yshayy:
To be fair, the main changes I see here are style changes, prettier was here before and with different settings. I don't know to conclude what is better, 2 or 4 spaces, or space between brackets.
I prefer that we use the exact same style before so we can review the real additonal value the eslint brings here.

Yshayy:
It looks like there are some new files as well

Yshayy:
Oh, I see it due to file name changes

Yshayy:
+1 For the file casing convention.

tomeresk:
> To be fair, the main changes I see here are style changes, prettier was here before and with different settings. I don't know to conclude what is better, 2 or 4 spaces, or space between brackets.
> I prefer that we use the exact same style before so we can review the real additonal value the eslint brings here.
> Right now, it's really difficult to find it.

Most of those style changes are in the `prettier format run` commit, in `Files changed` you can select all commits except that one (or just the last 18 after it) and then you see all of the non-style changes conveniently, what remains is not very long

Yshayy:
Reviewing it commit by commit is convenient. I think that other than the any->unknown and the no-null rule, everything looks great.
As for prettier, I really don't care for which style, but I think it can be in different PR.

.prettierrc
```diff
@@ -1,7 +1,4 @@
 {
```
AvivRubys, line 1: Why change these? I do like how they are now, particularly the 4 spaces

Yshayy:
I see that no-explicit-any is turned off, so that's good enough

.prettierrc
AleF83: I prefer 2 because 4 makes lines very long...
Let vote! 
👍  for 4
👎  for 2

services/src/modules/resource-repository/s3.ts
```diff
@@ -184,6 +184,7 @@ export class S3ResourceRepository implements ResourceRepository {
   }
 }
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
 function isAwsError(e: any): e is AWS.AWSError {
```
tomeresk, line 188: Since this function receives the value of the `catch` clause, even if we don't know the exact type of the error we do know it is derived from the `Error` class (all other error types should inherit from it).
We can use `Error` instead of `any` here

services/src/modules/shutdown-handler.ts
```diff
@@ -4,14 +4,15 @@ import * as process from 'process';
...
     console.log('Uncaught exception caught in global handler', error);
     // logger.fatal({error}, 'Uncaught exception caught in global handler');
     process.exit(1);
   });
 
-  process.on('unhandledRejection', (error: any) => {
-    console.log('Uncaught exception caught in global handler', error);
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  process.on('unhandledRejection', (reason: {} | null | undefined) => {
+    console.log('Uncaught exception caught in global handler', reason);
```
tomeresk, line 15: Should probably be `console.error` instead of `console.log`.
Same is true for the `uncaughtException` one

services/src/tests/e2e/tests/__snapshots__/authorization-with-queries.spec.ts.snap
```diff
@@ -10,8 +10,8 @@ Object {
         "exception": Object {
           "stacktrace": Array [
             "Error: Unauthorized by policy alwaysDenied in namespace namespace",
-            "    at Function.validatePolicy (/service/dist/modules/directives/policy/policy-executor.js:27:19)",
-            "    at async field.resolve (/service/dist/modules/directives/policy/policy.js:13:17)",
+            "    at Function.validatePolicy (/service/dist/modules/directives/policy/policy-executor.js:28:19)",
```
tomeresk, line 13: Not in this PR, but we should do something about the snapshots.
It's unreasonable to have to change this for any internal code change that has no effect on the outcome.

Do the snapshots support partial fuzzy matching for sections of the result?

services/src/tests/integration/gateway/rest-directive.spec.ts
```diff
@@ -77,5 +76,5 @@ function mockRestBackend(host: string) {
     .reply(200, 'world!')
     .get('/hello')
     .query({ name: 'miriam' })
-    .reply(200, (url) => parseUrl(url, true).query.name + '!');
+    .reply(200, (url) => new URL(url, host).searchParams.get('name') + '!');
```
tomeresk, line 79: Might as well use a Template literal instead of the `+`

services/src/tests/integration/registry/create-resources.spec.ts
```diff
@@ -17,7 +17,7 @@ import {
   UpstreamClientCredentials,
 } from '../../../modules/resource-repository/types';
 import { tmpPoliciesDir } from '../../../modules/config';
-import mockFsForOpa from '../../helpers/mockFsForOpa';
+import * as mockFsForOpa from '../../helpers/mock-fs-for-opa';
```
tomeresk, line 20: Same as before here, why remove the usage of the default export?

services/src/tests/integration/registry/update-resources.spec.ts
```diff
@@ -17,7 +17,7 @@ import {
   UpstreamClientCredentials,
 } from '../../../modules/resource-repository/types';
 import { tmpPoliciesDir } from '../../../modules/config';
-import mockFsForOpa from '../../helpers/mockFsForOpa';
+import * as mockFsForOpa from '../../helpers/mock-fs-for-opa';
```
tomeresk, line 20: Same as before here, why remove the usage of the default export?

tomeresk:
> I see that no-explicit-any is turned off, so that's good enough

Note that it is turned off only in test files, it is turned on in all the normal code files

services/src/tests/e2e/jest-e2e-environment.js
```diff
@@ -1,16 +1,16 @@
 const NodeEnvironment = require('jest-environment-node');
 const dockerCompose = require('docker-compose');
-const waitFor = require('./waitFor');
+const waitFor = require('./wait-for');
 
-const sleep = timeout => new Promise(r => setTimeout(r, timeout));
+const sleep = (timeout) => new Promise((r) => setTimeout(r, timeout));
```
tomeresk, line 5: Why do we require parentheses for a single param? it's kind of annoying and clutters the code

.prettierrc
tomeresk: I also prefer 2, with 4 spaces even 2-3 levels of nesting makes pretty much every line too long, especially with descriptive param/function names

Yshayy:
> that it is turned off only in test files, it is turned on in all the normal

So let's remove it on all files. Developers should be able to use "any" explicitly without needing to add "ts-ignore"

.prettierrc
Yshayy: I see no reason to change it (so, implicitly it means 4).
Either way, this change can be in a different PR as it's just a matter of style and not code quality.

Yshayy: In general, I'm pretty natural around this and don't see it as important, so if it's matter to you that much, change it to 2 (again, preferably in a different PR)

services/src/tests/e2e/tests/__snapshots__/authorization-with-queries.spec.ts.snap
AvivRubys: Who cares? They are autogenerated, just run em and they fix themselves
It's not ideal but it's really not worth investing time into IMO

tomeresk: How do you generate them? I manually had to copy the changes from the error about snapshot mismatch and replace it in the snapshot. I didn't see anything about auto-generating them

AvivRubys: https://jestjs.io/docs/en/snapshot-testing#updating-snapshots

services/src/modules/graphql-service.ts
```diff
@@ -49,6 +48,9 @@ export function createGraphQLService(config: { resourceGroups: Observable<Resour
 
       return sub.unsubscribe.bind(sub);
     },
+    executor<TContext>(requestContext: GraphQLRequestContextExecutionDidStart<TContext>) {
```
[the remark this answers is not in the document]
AleF83: The type of gateway argument in ApolloServer constructor was changed

services/src/modules/opa-helper.spec.ts
```diff
@@ -1,10 +1,10 @@
-import { getTmpFilePaths, prepareCompiledRegoFile, normalizeRegoCodePackage } from './opaHelper';
-import { mocked } from 'ts-jest/utils';
 import { promises as fs } from 'fs';
 import { exec } from 'child_process';
-import { tmpPoliciesDir } from './config';
 import * as path from 'path';
-import mockFsForOpa from '../tests/helpers/mockFsForOpa';
+import { mocked } from 'ts-jest/utils';
+import * as mockFsForOpa from '../tests/helpers/mock-fs-for-opa';
```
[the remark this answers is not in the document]
AleF83: It isn't default export. The mock-fs-for-opa hasn't default export. It exports 2 functions

services/src/modules/resource-repository/s3.ts
AleF83: Not really. In function body there are checks on `e.something` properties that aren't Error's properties. So we can or to use any or to create new variable with casting. In this case any is better IMO. Also all user-defined type guard examples are with any.

services/src/modules/shutdown-handler.ts
AleF83: It doesn't related to PR. But I'll change

services/src/tests/e2e/tests/__snapshots__/authorization-with-queries.spec.ts.snap
AleF83: I'll change the snapshot not to include stack trace in another PR

services/src/tests/integration/registry/create-resources.spec.ts
AleF83: The same comment as below :-)

services/src/tests/integration/registry/update-resources.spec.ts
AleF83: The same comment as below :-)

services/src/tests/e2e/jest-e2e-environment.js
AleF83: It's prettier default. I didn't do anything about this. I'll see if it can be changed

AleF83:
> > that it is turned off only in test files, it is turned on in all the normal
> 
> So let's remove it on all files. Developers should be able to use "any" explicitly without needing to add "ts-ignore"

It's warning for now. But I'll turn it off.

services/src/tests/e2e/jest-e2e-environment.js
AleF83: Ok. Found: https://prettier.io/docs/en/options.html#arrow-function-parentheses
Fixed

.prettierrc
AvivRubys: Whatever 🤷 not gonna argue about this

services/src/modules/opa-helper.spec.ts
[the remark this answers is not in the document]
tomeresk: It has a default export that exposes the 2 functions: `export default { mock, restore }`, and it was being used before

This is so you can import it like it was before (`import mockFsForOpa from '../tests/helpers/mockFsForOpa'`) instead of having to use `* as` like after the change.
I still don't see a good reason not to use the default

PR #150, opened 2020-07-07
Authorization documentation

AleF83:


docs/authorization.md
```diff
@@ -0,0 +1,121 @@
...
+| -------- | ----------------------------------------------------------------------------------------------------------------------------------------------- | --------- |
+| metadata | Contains two fields name and namespace.</br> Namespace is a logical group of policies.</br> Namespace and name pair is policy unique identifier | true      |
+| type     | Policy type.</br> Currently only [OPA](https://www.openpolicyagent.org) policies are supported                                                  | true      |
+| code     | Policy body.</br>Currently only [Rego](https://www.openpolicyagent.org/docs/latest/policy-language) language is supported.</br>                 | true      |
+| args     | Argument definitions.</br> These arguments is available in the policy body.                                                                     | false     |
+| query    | Graphql query to provide additional data to policy.</br> Its result is available in the policy body.                                            | false     |
+
+Example of policy definition:
```
tomeresk, line 19: I think this example will be overwhelming for anyone reading the docs that is unfamiliar with the feature.

A few notes about examples:

- I prefer the approach taken in the spec of starting with very simple examples using only the most basic of features and then using more examples to each introduce and showcase an additional feature. IMO these examples would be the most important part of the documentation for a feature like this.

- Examples are clearer when they use some obvious context for the namespace and name, like users, movies, accounts etc. (the context doesn't need to actually be defined anywhere).
For example, instead of `my_namespace` and `policy1` you can use the namespace `users` and policy name 'onlyAdmins'.

- I think we should have an example that clearly showcases the use of JWT in the policy, as that can potentially be a common use case

docs/authorization.md
```diff
@@ -0,0 +1,121 @@
...
+code: |
+  default allow = false
+  allow {
+    input.args.roles[_] == "admin"
+  }
+args:
+  roles: [String!]
+```
+
+In the example above the `roles` argument value is received using Argument Injection mechanism. In this case it is got from `roles` claim from request JWT.
```
tomeresk, line 115:
```suggestion
In the example above, the `roles` argument value is received by using the Argument Injection mechanism. In this example, it is received from the `roles` claim in the request's JWT.
```

docs/authorization.md
```diff
@@ -0,0 +1,121 @@
...
+
+> Note: See [Arguments Injection](./arguments_injection.md) for more information about configuration of policy arguments.
+
+## Limitations
+
+Currently only the access to Query type and it subtypes can be restricted by policies.
```
tomeresk, line 121: This is a little confusing, I assume you mean as opposed to Mutations?

AleF83: > This is a little confusing, I assume you mean as opposed to Mutations?

Yep. Currently we don't have authorization solution for Mutations.

docs/authorization.md
```diff
@@ -0,0 +1,149 @@
...
+
+```yaml
+metadata:
+  namespace: admin_ns
+  name: adminOnly
+type: opa
+code: |
+  default allow = false
+  allow {
+    input.args.roles[_] == "admin"
```
tomeresk, line 109: The roles actually come from the query here (and not args)

```suggestion
    input.query.user.roles[_] == "admin"
```

docs/authorization.md
```diff
@@ -0,0 +1,149 @@
...
+## Policy directive
+
+The policies are set using graphql directives. In order to attach policy to field the `@policy` directive should be set on the field.
+So if in type `User` there is field `phone` that should be accessible for user with admin role only the type definition should be as following:
+
+```graphql
+type User {
+  id: ID!
+  name: String!
+  phone: String @policy(namespace: "admin_ns", name: "admin_only", args: { roles: "{jwt.roles}" })
```
tomeresk, line 141: Since you mentioned this is for the policy defined in example 1, we should probably stay consistent with the naming there:

```suggestion
  phone: String @policy(namespace: "billing", name: "adminOnly", args: { userRoles: "{jwt.roles}" })
```

PR #151, opened 2020-07-07
Arguments Injection

AleF83:


services/src/modules/arguments-injection/arguments-injection.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+
+interface TestCase {
+  template: string;
+  source?: unknown;
+  args?: Record<string, unknown>;
+  context?: RequestContext;
+  expected: unknown;
+}
+
+const testCases: [string, TestCase][] = [
```
tomeresk, line 12: Probably worth adding a test case for `jwt` since it has special handling (being taken from the `context`). Also possibly for `exports` (though that one might be tricky)

PR #152, opened 2020-07-12
Authorization @policy directive on object type

AleF83:


services/src/modules/build-federated-schema.ts
```diff
@@ -69,8 +73,17 @@ export function buildSchemaFromFederatedTypeDefs({
 
   // Add directives back to the SDL
   const fullTypeDefWithoutDirectives = parse(printSchema(compositionResult.schema));
-  const allDirectivesUsages = mergeDirectivesUsages(serviceDefinitions.map(sd => sd.directivesUsages));
-  const fullTypeDefWithDirectives = addDirectivesToTypedefs(allDirectivesUsages, fullTypeDefWithoutDirectives);
+  const objectTypeDirectives = mergeObjectTypeDirectivesUsages(
```
AvivRubys, line 76: Make sure these handle nulls well i.e. a servicedefinition with object defs and no field defs, etc

services/src/modules/directives/policy/policy.ts
```diff
@@ -1,30 +1,69 @@
-import { GraphQLResolveInfo, GraphQLField, defaultFieldResolver } from 'graphql';
+import {
```
AvivRubys, line 1: I can't really understand the changes in this file. What's going on here?

services/src/tests/e2e/utils/graphql-error-serializer.ts
```diff
@@ -0,0 +1,19 @@
+import { GraphQLError } from 'graphql';
```
AvivRubys, line 1: Nice solution for this 👍

tomeresk: 👍

services/src/modules/directives/policy/policy.ts
AleF83: I've added commit with comments

services/src/modules/build-federated-schema.ts
AleF83: All variables has default value of `{}`. So it should be ok.

services/src/modules/directives/policy/policy.ts
AvivRubys: Ok, it boils down to wanting the policy to always be the first directive to run, if I understand correctly.
That's okay, but I think it really doesn't belong in the `policy` directive implementation.
I think we should either -
1. Separate directives into "pre" "resolver" and "post" directives, and then when the schema is built we can call `visitSchemaDirectives` multiple times to fit the reverse-order the directives should run in
1. Same idea, but use a per-resolver middleware style solution like the agogos implementation
1. Extract this code out of here and make it into a kind of higher-order resolver function we call from this file

In general I always think if you can solve the issue through design rather than through hacks like `Proxy` that's a better idea.

services/src/modules/build-federated-schema.ts
AvivRubys: Awsome, I just wasn't 100% sure on this 👍

services/src/modules/build-federated-schema.ts
```diff
@@ -108,43 +137,54 @@ function addDirectivesToTypedefs(directives: DirectivesUsagesByObjectAndFieldNam
 const federationDirectives = ['key', 'extends', 'external', 'requires', 'provides'];
 
 export function collectAndRemoveCustomDirectives(typeDef: DocumentNode) {
-  const directivesUsages: DirectivesUsagesByObjectAndFieldNames = {};
+  const objectTypeDirectivesUsages: DirectivesUsagesByObjectName = {};
+  const fieldDefinitionDirectivesUsages: DirectivesUsagesByObjectAndFieldNames = {};
 
   const typeDefWithoutDirectives = visit(typeDef, {
     Directive(node, _key, _parent, _path, ancestors) {
```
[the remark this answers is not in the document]
AleF83: I've tried to split but it doesn't look good. The `collectAndRemoveCustomDirectives` function body is mostly the `visit` function with visitor function. And this function mostly alters the directive dictionaries..

services/src/modules/directives/policy/policy.ts
AleF83: In general I agree and the initial approach was alike :-) But I afraid it cannot be implemented here.

We have 2 types of directives now.
1) replaces the field resolver
2) wraps the field resolver
In addition to this, Apollo server extension mechanism wraps everything one more time.
One more thing, the directives can be attached to object type and field definition.

For example `policy` directive should be applied after the `select` directive on field but both directives are wrapping directives. So both should be "post-resolve". So now we need to order inside "post-resolve" group.
The object directives are applied before the field definition ones but `policy` directive on object should be applied after `select` directive on field definition.

I think we can stay with this hack for `policy` directive until we'll encounter another one that should be threatened alike.

tomeresk: I would also be happy if we find a design solution that removes the need for this hack, but we should continue this discussion elsewhere rather than on this PR (and merge the PR with this hack for now).

Any such decision would be a big change (at least conceptually) with future implications on directives, so we should probably discuss it without stalling this PR

PR #153, opened 2020-07-14
Policy query variables can depend only on `args`

AleF83:


docs/authorization.md
```diff
@@ -150,4 +150,7 @@ In the example above, the `roles` argument value is received by using the Argume
 
 ## Directives order
 
-**_Important!_** The [order of directives](https://github.com/graphql/graphql-spec/blob/master/spec/Section%202%20--%20Language.md#directives) does matter! In the most cases the `@policy` directive should be **the last** directive attached to the field definition.
+**_Important!_** The [order of directives](https://github.com/graphql/graphql-spec/blob/master/spec/Section%202%20--%20Language.md#directives) does matter! The directives are applied in order they appear in schema. Object type directives are applied before field definition directives.
+
+The `@policy` directive (that can be set either on object type or field definition) is implemented so that the policy check will run after all other directive resolvers.
```
tomeresk, line 155: This is a little confusing since the policy directive runs _after_ the other resolvers meaning the policy check actually runs _before_ the other directives. Maybe change the wording here to be less explicit but less confusing:

```suggestion
The `@policy` directive (that can be set either on object type or field definition) is implemented so that the policy requirements are tested before all other directive resolvers, regardless of the order of directives.
```

PR #155, opened 2020-07-14
Authorization - Policy caching

tomeresk:
Best review each commit separately.
The first commit lays down some groundwork by refactoring PolicyExecutor to have one shared instance per incoming request, and the second commit actually adds the caching.

Regarding the `CachedOperation` abstract class, I originally wanted it to be a Mixin.
I eventually went with the abstract class since Typescript has no proper support for Mixins with generics (The [ts-mixer module has a solution for this](https://github.com/tannerntannern/ts-mixer#mixing-generic-classes), but it's also kind of hacky)


services/src/tests/integration/gateway/policy-caching.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+            firstName
+            middleName
+            lastName
+          }
+        }
+      `,
+    });
+
+    expect(response.errors).toBeUndefined();
+    expect(response.data).toEqual({ user: { id: '42', firstName: 'John', middleName: 'Hope', lastName: 'Smith' } });
```
AleF83, line 104: You can use snapshot

AleF83 (approved):

services/src/tests/integration/gateway/policy-caching.spec.ts
tomeresk: Seems excessive when the data is this short

PR #157, opened 2020-07-15
Use 'wait-on' module to wait for docker-compose

AleF83:


AleF83:
> I don't really understand why do you waste your time with changes like these though 🤷 it was already working

Some kind of OCD..

PR #159, opened 2020-07-19
Use resolveObject feature for policy directive on object type

AleF83:
This is `policy` directive implementation using `resolveObject` experimental feature of Apollo.
It should be introduced officially in next major version.


services/tests/integration/gateway/policy-directive.spec.ts
```diff
@@ -219,6 +222,7 @@ const testCases: [string, TestCase][] = [
   [
     'Stub, policy, lowerCase on field (ALLOW)',
     {
+      // This is incorrect order of directives but it still does work by chance.
```
tomeresk, line 225: I don't think `by chance` is the correct term here, it still works because the `lowerCase` directive specifically calls the next resolver, but might not work with another directive after it

services/tests/integration/gateway/policy-directive.spec.ts
```diff
@@ -230,6 +234,7 @@ const testCases: [string, TestCase][] = [
     },
   ],
   [
+    // This is incorrect order of directives but it still does work by chance.
```
tomeresk, line 237: same here

PR #166, opened 2020-07-27
Policy attachment changes

tomeresk:
Don't load policy attachments cache when on registry
Save the loaded WASM instances in the policy attachments cache instead of the buffer of the WASM code

### DIFF
--- a/services/src/modules/directives/policy/types.ts
+++ b/services/src/modules/directives/policy/types.ts
@@ -54,3 +54,8 @@ export type PolicyCacheKey = {
   metadata: ResourceMetadata;
   args?: PolicyArgsObject;
 };
+
+// Temporary until opa-wasm package adds types
+export interface LoadedPolicy {
+  evaluate: (input: { [key: string]: any }) => any[];
+}
--- a/services/src/modules/directives/policy/types.ts
+++ b/services/src/modules/directives/policy/types.ts
@@ -57,5 +57,5 @@ export type PolicyCacheKey = {
 
 // Temporary until opa-wasm package adds types
 export interface LoadedPolicy {
-  evaluate: (input: { [key: string]: any }) => any[];
+  evaluate: (input: Record<string, unknown>) => any[];
 }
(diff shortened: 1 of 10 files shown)


services/src/modules/directives/policy/types.ts
```diff
@@ -54,3 +54,8 @@ export type PolicyCacheKey = {
   metadata: ResourceMetadata;
   args?: PolicyArgsObject;
 };
+
+// Temporary until opa-wasm package adds types
+export interface LoadedPolicy {
+  evaluate: (input: { [key: string]: any }) => any[];
```
AleF83, line 60:
```suggestion
  evaluate: (input: Record<string, unknown>) => { result: { allow: boolean } }[];
```

services/src/registry.ts
```diff
@@ -180,7 +180,7 @@ interface PolicyInput {
   query?: PolicyQueryInput;
 }
 
-const resourceRepository = S3ResourceRepository.fromEnvironment();
+const resourceRepository = S3ResourceRepository.fromEnvironment({ isRegistry: true });
```
AleF83, line 183: `isRegistry` flag looks very general. Maybe we can use specific flag (`disablePoliciesLoading` or something like this). This flag can be initialized in `env-var` to default `false` and set in `registry` to `true`.

AleF83 (changes requested):

services/src/registry.ts
tomeresk: I don't see why we would want a specific flag here, it just hides the reason we decided to do it (which is that the code runs in the registry).

If we don't separate code between the registry and the gateway we do need a way to tell between them when necessary.
I prefer letting each function decide the necessary logic based on the information it needs (whether we are in the registry or not) than the registry and gateway having intimate knowledge about each feature and deciding which flags to send for every feature.

services/src/modules/directives/policy/types.ts
tomeresk: The `LoadedPolicy` is supposed to be a generic type for a loaded WASM policy, to use until opa-wasm module is typed and replaced with the new type later.

I'll change it to `Record`, but the type of the result is a convention of our policy and not something for the generic loaded WASM policy type

AleF83 (approved):

PR #167, opened 2020-08-03
cli - add support for policies

tomeresk:
Note the `cli/src/client/types.ts` and `cli/oclif.manifest.json` files are auto-generated


cli/src/client/types.ts
```diff
@@ -1,46 +1,72 @@
 // Autogenerated from https://graphql-code-generator.com/#live-demo
 
 export type Maybe<T> = T | null;
+export type Exact<T extends { [key: string]: unknown }> = { [K in keyof T]: T[K] };
```
AleF83, line 4: I don't see this in use... Also, there is Pick<> type..

tomeresk: As I wrote in the description, this file is Auto-generated, I prefer not to make changes there since they will be overridden on any future change (When it will be re-generated again)

AleF83 (approved):

PR #168, opened 2020-08-03
Authentication

AleF83:
This PR adds authentication to Stitch.
There is support for different authentication strategies.
Currently two strategies are implemented:
1. JWT provided in Authorization header
2. Anonymous access to defined paths.


services/src/modules/authentication/get-secret.ts
```diff
@@ -0,0 +1,49 @@
+import fastify from 'fastify';
+
+// There is problem with typings in the module: https://github.com/auth0/node-jwks-rsa/issues/78
+import * as jwks from 'jwks-rsa';
+import { ClientOptions, JwksClient } from 'jwks-rsa';
+import { decode } from 'jsonwebtoken';
+import logger from '../logger';
+import { authenticationConfig } from '../config';
+
+type DecodeJWT = {
```
tomeresk, line 10: Should probably be `DecodedJWT`

services/src/modules/authentication/get-secret.ts
```diff
@@ -0,0 +1,49 @@
...
+const createClient = (jwks as unknown) as (options: ClientOptions) => JwksClient;
+
+export default function getSecret(
+  request: fastify.FastifyRequest,
+  _reply: fastify.FastifyReply<unknown>,
+  cb: (e: Error | null, secret: string | undefined) => void
+): void {
+  const authHeader: string = request.headers.authorization;
+  const token = authHeader.split(' ')[1];
+  const decodedToken = decode(token, { complete: true, json: true }) as DecodeJWT;
```
tomeresk, line 25: Can't we somehow re-use the token in the [request context](https://github.com/Soluto/stitch/blob/master/services/src/modules/gateway.ts#L27) (or use the one decoded here in there) instead of decoding it twice in both places?

services/src/modules/authentication/strategies/anonymous.ts
```diff
@@ -0,0 +1,15 @@
+import fastify from 'fastify';
+import { authenticationConfig } from '../../config';
+
+export default function (
+  request: fastify.FastifyRequest,
+  _reply: fastify.FastifyReply<unknown>,
+  done: (error?: Error) => void
+) {
+  const anonymousPaths = authenticationConfig?.anonymous?.paths;
+  if (anonymousPaths && anonymousPaths.some(ap => request.raw.url?.endsWith(ap))) {
```
tomeresk, line 10:
```suggestion
  if (anonymousPaths?.some(ap => request.raw.url?.endsWith(ap))) {
```

services/src/modules/authentication/strategies/jwt.ts
```diff
@@ -0,0 +1,25 @@
...
+  if (!issuerConfig) {
+    logger.warn({ issuer }, 'Unknown issuer');
+    return Promise.reject(new Error('Unauthorized'));
+  }
+  if (issuerConfig.authenticatedPaths.some(ap => request.raw.url?.endsWith(ap))) {
+    try {
+      await request.jwtVerify();
+    } catch (error) {
+      logger.warn(error, 'Failed to verify request JWT');
+      return Promise.reject(new Error('Unauthorized'));
```
tomeresk, line 22: It will automatically wrap it in a rejected Promise since we are in an `async` function

```suggestion
      throw new Error('Unauthorized');
```

services/src/modules/authentication/strategies/jwt.ts
```diff
@@ -0,0 +1,25 @@
+import fastify from 'fastify';
+import { decode } from 'jsonwebtoken';
+import logger from '../../logger';
+import { authenticationConfig } from '../../config';
+
+export default async function (request: fastify.FastifyRequest): Promise<void> {
+  const authHeader: string = request.headers.authorization;
+  if (!authHeader) return Promise.reject(new Error('Unauthorized'));
+  const token = authHeader.split(' ')[1];
+  const decodedToken = decode(token, { json: true }) as Record<string, unknown>;
```
tomeresk, line 10: Third place we decode the token for the same request, by this point we must just put the decoded token on the fastify request (which is available in all places) and decode only if we can't find it on the request (so only the first one will actually decode it).
Probably just make a helper function somewhere that gets a fastify requests and returns the decoded token (it will know to check if it was already previously decoded and saved on the request, so the other places won't have to know the structure of the request context)

services/src/modules/config.ts
```diff
@@ -1,5 +1,7 @@
 import * as path from 'path';
 import * as envVar from 'env-var';
+import { AuthenticationConfig } from './authentication/types';
+// import { AuthenticationConfig } from './authentication/types';
```
tomeresk, line 4: Can be removed?

services/src/modules/config.ts
```diff
@@ -23,3 +25,6 @@ export const tmpPoliciesDir = envVar
   .get('TMP_POLICIES_DIR')
   .default(path.resolve(process.cwd(), 'tmp/policies'))
   .asString();
+
+// Authentication
+export const authenticationConfig = envVar.get('AUTHENTICATION_CONFIGURATION').asJsonObject() as AuthenticationConfig;
```
tomeresk, line 30: We should probably have some simple default that works for dev env/tests

services/tests/e2e/authorization/auth-jwt.spec.ts
```diff
@@ -8,14 +7,18 @@ import {
...
+const gatewayClient = new GraphQLClient('http://localhost:8080/graphql');
+const registryClient = new GraphQLClient('http://localhost:8090/graphql');
 
 describe('authorization', () => {
-  beforeAll(() => {
+  beforeAll(async () => {
     expect.addSnapshotSerializer(GraphQLErrorSerializer);
+
+    const accessToken = await getToken();
+    gatewayClient.setHeader('Authorization', `Bearer ${accessToken}`);
```
tomeresk, line 21: We should probably have this in `beforeEach` since some tests override this with a specific token, and we want to restore the standard one for the next tests.
We can generate the token only once in `beforeAll` and then just set it every time in `beforeEach`

tomeresk (changes requested):

services/src/modules/authentication/strategies/jwt.ts
AleF83: Yeah. I noticed this. I'll add plugin to fastify that will set decoded jwt value on the request (if JWT exists in authorization header).

services/src/modules/config.ts
```diff
@@ -27,4 +26,10 @@ export const tmpPoliciesDir = envVar
...
 
 // Authentication
-export const authenticationConfig = envVar.get('AUTHENTICATION_CONFIGURATION').asJsonObject() as AuthenticationConfig;
+const defaultAuthenticationConfig: AuthenticationConfig = {
+  anonymous: {
+    paths: ['/metrics', '/.well-known/apollo/server-health', '/graphql'],
+  },
+};
+export const authenticationConfig =
+  (envVar.get('AUTHENTICATION_CONFIGURATION').asJsonObject() as AuthenticationConfig) ?? defaultAuthenticationConfig;
```
tomeresk, line 35: I think this will also be fine with the `env-var` module default function?
`envVar.get('AUTHENTICATION_CONFIGURATION').default(defaultAuthenticationConfig).asJsonObject()`

tomeresk (approved):
Just the token decoding left, so I assume you meant to do that in another PR?

services/src/modules/authentication/jwt-decoder-plugin.ts
```diff
@@ -0,0 +1,43 @@
...
+    if (maybeDecodedJWT) {
+      this.decodedJWT = maybeDecodedJWT;
+    }
+  } catch (e) {
+    logger.warn(e, 'Fail to extract and decode JWT from request');
+  }
+}
+
+export default fp(async function (instance: FastifyInstance) {
+  console.log('===== PLUGIN ENABLED =====');
```
tomeresk, line 41: Can probably remove this at this point

services/src/modules/authentication/jwt-decoder-plugin.ts
```diff
@@ -0,0 +1,43 @@
...
+    Params = DefaultParams,
+    Headers = DefaultHeaders,
+    Body = DefaultBody
+  > {
+    decodeJWT(): Promise<void>;
+    decodedJWT?: DecodedJWT | null;
+  }
+}
+
+async function decodeRequestJwt(this: FastifyRequest) {
```
tomeresk, line 26: It is very unusual to name an argument `this`, you should change it to something else

services/src/modules/authentication/jwt-decoder-plugin.ts
```diff
@@ -0,0 +1,43 @@
...
+  try {
+    const authHeader = this.headers.authorization;
+    if (!authHeader) return;
+    const authToken = String(authHeader).split(' ')[1];
+    const maybeDecodedJWT = (await decode(authToken, { complete: true, json: true })) as DecodedJWT | null;
```
tomeresk, line 31:
```suggestion
    const maybeDecodedJWT = (await decode(authToken, { complete: true, json: true })) as DecodedJWT || null;
```

services/src/modules/authentication/strategies/jwt.ts
```diff
@@ -1,14 +1,11 @@
...
 import logger from '../../logger';
 import { authenticationConfig } from '../../config';
 
 export default async function (request: fastify.FastifyRequest): Promise<void> {
-  const authHeader: string = request.headers.authorization;
-  if (!authHeader) throw new Error('Unauthorized');
-  const token = authHeader.split(' ')[1];
-  const decodedToken = decode(token, { json: true }) as Record<string, unknown>;
-  const issuer = String(decodedToken?.iss);
+  await request.decodeJWT();
```
tomeresk, line 6: I don't really like the fact that we have to know this place is the first to use the JWT and manually call decode, and the other places just have to hope someone else already decoded it.

Two possible solutions:

1. Change `decodeJWT` function to first check if `decodedJWT` is already set, and do nothing if it is. Then, change every place that uses JWT to first call `decodeJWT` and then use it.

2. Turn `decodedJWT` into a proxy with a getter, where the getter actually does the decoding on the first invocation and simply returns the decoded token on the next invocations. Every place that needs to use `decodedJWT` simply uses it without needing to know if they are first or some place else previously decoded it.

option 2 is probably the better option, but takes a little more work

tomeresk (changes requested):

services/src/modules/authentication/jwt-decoder-plugin.ts
AleF83: This is not an argument.
See [here](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#specifying-the-type-of-this-for-functions).

services/src/modules/authentication/jwt-decoder-plugin.ts
AleF83: `decode` [returns](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/jsonwebtoken/index.d.ts#L209) `null | { [key: string]: any }`. I just wanted to make the second part more detailed. The single pipe is here intentionally.

services/src/modules/authentication/strategies/jwt.ts
AleF83: Agree. I'll implement the first option for now.

services/src/modules/authentication/jwt-decoder-plugin.ts
```diff
@@ -18,26 +18,29 @@ declare module 'fastify' {
...
     if (!authHeader) return;
     const authToken = String(authHeader).split(' ')[1];
-    const maybeDecodedJWT = (await decode(authToken, { complete: true, json: true })) as DecodedJWT | null;
+    const maybeDecodedJWT = decode(authToken, { complete: true, json: true }) as DecodedJWT | null;
     if (maybeDecodedJWT) {
-      this.decodedJWT = maybeDecodedJWT;
+      this._decodedJWT = maybeDecodedJWT;
+      return this._decodedJWT;
     }
+    return;
```
tomeresk, line 37: This `return` is meaningless right now

services/src/modules/authentication/jwt-decoder-plugin.ts
```diff
@@ -18,26 +18,29 @@ declare module 'fastify' {
...
   } catch (e) {
     logger.warn(e, 'Fail to extract and decode JWT from request');
+    return;
```
tomeresk, line 40: this `return` is also meaningless

tomeresk (approved):

PR #179, opened 2020-08-19
@mergeWithStub

tomeresk:


services/src/modules/directives/merge-with-stub.ts
```diff
@@ -0,0 +1,23 @@
...
+export class MergeWithStubDirective extends SchemaDirectiveVisitor {
+  visitFieldDefinition(field: GraphQLField<unknown, RequestContext>) {
+    const { value } = this.args;
+    const originalResolve = field.resolve || defaultFieldResolver;
+
+    field.resolve = async (parent, args, context, info) => {
+      const originalValue = await originalResolve.call(field, parent, args, context, info);
+      const stubValue = typeof value === 'string' ? inject(value, parent, args, context, info) : value;
+
+      return { ...originalValue, ...stubValue };
```
AleF83, line 16: Maybe we can add optional parameter to the directive named `deepMerge` of type boolean with default false and if it's true use lodash or ramda merge...

AleF83 (approved):

services/src/modules/directives/merge-with-stub.ts
tomeresk: Sounds reasonable, but it won't be a breaking change so we can wait until we need it (since I'm not 100% sure there will be a use case for it)

docs/data_sources.md
```diff
@@ -112,6 +112,22 @@ Full parameters:
...
+
+```graphql
+type User {
+    firstName: String!
+    lastName: String!
+    fullName: String // not supplied by the resolver
+}
+
+type Query {
+    getUser: User! @mergeWithStub(value: "{{ fullName: `${source.firstName} ${source.lastName}` }}")
```
AvivRubys, line 127: Kind of a nitpick, but probably better to promote using the `JSON` scalar here instead of a string
```suggestion
    getUser: User! @mergeWithStub(value: {fullName: "${source.firstName} ${source.lastName}"})
```

docs/data_sources.md
```diff
@@ -112,6 +112,22 @@ Full parameters:
 directive @stub(value: JSON!)
 ```
 
+## @mergeWithStub
```
AvivRubys, line 115: I'm not crazy about the name... I'm spitballing a bit here, but what if this directive was implemented as a parameter to `@stub`?
So you could do something like `field: SomeEntity! @stub(value: {asd: 70}, strategy: REPLACE/MERGE)` (Strategy being an enum, could also have DEEPMERGE to address @AleF83 's concerns)
Maybe renaming `@stub` could go along with this, to `@inline` or `@static`or something

docs/data_sources.md
tomeresk: I kept the injection implementation the same as stub, so this won't work on either of them (they only inject for string, anything else is returned as is)

Since I don't really see a reason not to, I'll just change both to also support deep injection (in addition to making this docs change)

docs/data_sources.md
tomeresk: When we initially discussed making this directive we did consider just doing it as a parameter on `@stub`, but we dropped the idea because it sounds weird with the `stub` name, and changing the `stub` name would be a breaking change so it would complicate things and take more time.

We could just rename the `mergeWithStub` directive and absorb the stub functionality into it (with the `strategy` parameter), while marking the `@stub` directive as deprecated (and we can remove it at a later point). WDYT?

services/src/modules/directives/local-resolver.ts
```diff
@@ -0,0 +1,46 @@
...
+          ? inject(value, parent, args, context, info)
+          : deepInject(value, parent, args, context, info);
+
+      if (mergeStrategy === MergeStrategy.Replace) return stubValue;
+
+      const originalValue = await originalResolve.call(field, parent, args, context, info);
+
+      return mergeStrategy === MergeStrategy.MergeDeep
+        ? lodash.merge(originalValue, stubValue)
+        : { ...originalValue, ...(stubValue as any) };
```
AleF83, line 25:
```suggestion
        : { ...originalValue, ...(stubValue as Record<string, unknown>) };
```

AleF83 (approved):

services/src/modules/directives/local-resolver.ts
```diff
@@ -0,0 +1,46 @@
...
+import { inject, deepInject } from '../arguments-injection';
+import { RequestContext } from '../context';
+
+export class LocalResolverDirective extends SchemaDirectiveVisitor {
+  visitFieldDefinition(field: GraphQLField<unknown, RequestContext>) {
+    const { value, mergeStrategy } = this.args as LocalResolverArgs;
+    const originalResolve = field.resolve || defaultFieldResolver;
+
+    field.resolve = async (parent, args, context, info) => {
+      const stubValue =
```
AvivRubys, line 14: `value` could be any JSON value - this covers string specifically, but what would happen with number/bool/array? Can `deepInject` handle those? It's fine if we don't want to handle it, but it's legal input and it might crash something here. Might be worth a unit test

AvivRubys: Same for line 25 where you spread the result of the injection - need to be careful there.

tomeresk: > `value` could be any JSON value - this covers string specifically, but what would happen with number/bool/array? Can `deepInject` handle those? It's fine if we don't want to handle it, but it's legal input and it might crash something here. Might be worth a unit test

The changes I made to `deepInject` were so that it could handle those cases as well (it returns those values as is)


> Same for line 25 where you spread the result of the injection - need to be careful there.

This can't cause issues that I know of, if you spread null/undefined/number or anything that isn't iterable it just won't affect the result (same as spreading an empty object), but the original value would still be correctly spread.
With iterables it will simply spread them (e.g. the string "aa" would be spread to `{ '0': 'a', '1': 'a' }`), so it also shouldn't cause issues.
I don't really see a reason to block this behavior with a check if someone wants to use it, it will also be easier to see if you sent a wrong input and it's spread this way than if it is simply ignored.

AvivRubys: Oh ok, I didn't look too deep into deepInject, if it can handle them that's fine 👍 
lgtm

AvivRubys (approved):

tomeresk:
@AleF83 @AvivRubys In the last commit I added validation to the stub value of `@localResolver`:
With the `Replace` strategy any stub value is fine
With the `Merge` and `MergeDeep` strategies the evaluated stub value must be an object or it throws an error

Additionally, I noticed `deepInject` didn't support arrays (both top level or if they were nested inside the object) so I ended up refactoring argument injection to only have a single `inject` function (instead of `inject` and `deepInject`) that handles any value, and recursively injects arrays/objects

AleF83 (approved):

PR #180, opened 2020-08-23
Logs & Metrics

AleF83:


services/src/modules/plugins/logging.ts
```diff
@@ -0,0 +1,55 @@
...
+  GraphQLRequestContextWillSendResponse,
+} from 'apollo-server-plugin-base';
+import { FastifyInstance } from 'fastify';
+import { Counter, Histogram } from 'prom-client';
+import logger from '../logger';
+
+let errorsInResponseCounter: Counter<string>;
+let requestDurationHistogram: Histogram<string>;
+
+export function createLoggingPlugin(fastifyInstance: FastifyInstance): ApolloServerPlugin {
```
AvivRubys, line 13: Maybe we can separate logging from metrics, either via different plugins or just more separation inside the function? Because right now `createLoggingPlugin`'s code is like 90% metrics

services/src/modules/plugins/logging.ts
```diff
@@ -0,0 +1,55 @@
...
+  return {
+    serverWillStart() {
+      const promClient = fastifyInstance.metrics?.client;
```
AvivRubys, line 16: Using `serverWillStart` to initialize metrics means that whenever theres a new ResourceGroup, the metrics will be re-initialized. I think they should be global instead, or scoped to the fastify app if that makes it easier for testing.

services/src/modules/plugins/logging.ts
```diff
@@ -0,0 +1,55 @@
...
+      errorsInResponseCounter = new promClient.Counter({
```
AvivRubys, line 17: Maybe we can do `graphql_result{status='success|error'}` or something like that instead?

services/src/modules/plugins/logging.ts
```diff
@@ -0,0 +1,55 @@
...
+          } = willSendResponseContext;
+
+          const endHrTime = process.hrtime.bigint();
+          const durationMs = Number(endHrTime - startHrTime) / 1000000;
+
+          endTimer();
+          const logData = { query, operationName, errors, durationMs };
+          if (errors) {
+            errorsInResponseCounter.inc();
+            logger.warn(logData, 'The server encountered errors while proceeding request');
```
AvivRubys, line 47: I don't think we should `warn` on every error. This means every upstream which has some error will make a warn log here, for example. Maybe reduce to `info`? I'm not sure.

services/src/modules/plugins/logging.ts
```diff
@@ -0,0 +1,55 @@
...
+      const endTimer = requestDurationHistogram.startTimer();
+      return {
+        willSendResponse(willSendResponseContext: GraphQLRequestContextWillSendResponse<unknown>) {
+          const {
+            request: { query, operationName },
+            response: { errors },
+          } = willSendResponseContext;
+
+          const endHrTime = process.hrtime.bigint();
+          const durationMs = Number(endHrTime - startHrTime) / 1000000;
```
AvivRubys, line 41: No reason to do this ourselves - it's built into [hrtime](https://nodejs.org/api/process.html#process_process_hrtime_time)

AvivRubys (changes requested):
What about instrumenting the resolvers themselves? I think something like that could be really useful. For example -
`graphql_resolver{object="User", field="UserId", result="sucess/error"}`
Maybe a timer too?

services/src/modules/plugins/logging.ts
AleF83: I considered it. I did both thing in one plugin because there is no a lot of code. But you're right. I'll separate them

services/src/modules/plugins/logging.ts
AleF83: The client is created in `fastify-metrics` plugin. I saw that metrics are incremented. I'll check it. Maybe existence check will be enough.

services/src/modules/plugins/logging.ts
AleF83: If it will be info and the log level will be `warn` we won't know about failures. But I'm not sure too.

services/src/modules/plugins/logging.ts
AleF83: I saw it but I didn't use it because the following sentence:
 `This is the legacy version of process.hrtime.bigint() before bigint was introduced in JavaScript.`

AleF83:
> What about instrumenting the resolvers themselves? I think something like that could be really useful. For example -
> `graphql_resolver{object="User", field="UserId", result="sucess/error"}`
> Maybe a timer too?

I didn't add it because there will be a lot of labels. But actually, it will be at most several hundred of labels so it won't crash prometheus. I'll add.

services/src/modules/plugins/metrics.ts
```diff
@@ -0,0 +1,56 @@
...
+      const endResponseTimer = requestDurationHistogram?.startTimer();
+      return {
+        willSendResponse(willSendResponseContext: GraphQLRequestContextWillSendResponse<unknown>) {
+          const {
+            response: { errors },
+          } = willSendResponseContext;
+          endResponseTimer?.({ status: errors ? 'error' : 'success' });
+        },
+        executionDidStart: () => ({
+          willResolveField(
```
AvivRubys, line 38: To reduce the number of labels, maybe should do this only for custom resolvers? We can check if `field.resolver === defaultFieldResolver`, and if so - don't measure it.

services/src/modules/plugins/metrics.ts
```diff
@@ -0,0 +1,56 @@
...
+  GraphQLRequestContextWillSendResponse,
+  GraphQLFieldResolverParams,
+} from 'apollo-server-plugin-base';
+import { FastifyInstance } from 'fastify';
+import { Histogram } from 'prom-client';
+
+let requestDurationHistogram: Histogram<string> | undefined;
+let resolverDurationHistogram: Histogram<string> | undefined;
+
+export function createMetricsPlugin(fastifyInstance: Pick<FastifyInstance, 'metrics'>): ApolloServerPlugin {
```
AvivRubys, line 12: If all we need is `metrics.client`, maybe just pass that bit in? Then the metrics plugin only depends on `prom-client`, not on fastify itself

AvivRubys (approved):
Two more small things, but I already like it much more now as it is 👍

services/src/modules/plugins/metrics.ts
AleF83: The problem that this client isn't created when the plugins are defined. This is the reason I send the instance itself

services/src/modules/plugins/metrics.ts
AleF83: I've added fix

PR #185, opened 2020-09-14
Support multiple @policy directive instances at the same location

AleF83:


tomeresk (approved):

AleF83:
This approach isn't clear enough about the relation between the @policy directives. (OR/AND)

PR #187, opened 2020-09-16
Argument Injection Builtin Module

AleF83:


services/src/modules/arguments-injection/arguments-evaluation.ts
```diff
@@ -1,11 +1,12 @@
 import { VM } from 'vm2';
+import { argumentInjectionBuiltIns } from '../config';
```
tomeresk, line 2: Don't really like the `BuiltIns` naming, I prefer `Helpers` (or `HelperFunctions` if you insist on being verbose)

tomeresk (commented):
LGTM, but requires very descriptive docs since the nature of the `BuiltIns` is very open-ended

AleF83:
Implemented in #190

PR #190, opened 2020-10-05
Plugins

AleF83:


services/src/modules/plugins/index.ts
```diff
@@ -0,0 +1,41 @@
+import * as fs from 'fs';
```
tomeresk, line 1: Probably better to just import the promises directly instead of explicitly accessing it later (`import { promises as fs } from 'fs';`), especially since in node 14 it's already a separate import (`import fs from 'fs/promises'`)

services/src/modules/plugins/index.ts
```diff
@@ -0,0 +1,41 @@
...
+  } catch (err) {
+    logger.error({ err }, 'Failed to build argument injection globals');
+    throw err;
+  }
+}
+
+export async function buildArgumentInjectionGlobals(): Promise<Record<string, unknown>> {
+  return plugins
+    .filter(p => p.addArgumentInjectionGlobals)
+    .reduce(async (acc, cur) => ({ ...acc, ...(await cur.addArgumentInjectionGlobals!()) }), {});
```
tomeresk, line 41: This can allow collisions with a later plugin overriding a value from a previous plugin.
We can try to avoid that by namespacing everything the plugin returns under the plugin name, not sure if we should though

services/src/modules/plugins/index.ts
```diff
@@ -39,3 +40,23 @@ export async function buildArgumentInjectionGlobals(): Promise<Record<string, un
     .filter(p => p.addArgumentInjectionGlobals)
     .reduce(async (acc, cur) => ({ ...acc, ...(await cur.addArgumentInjectionGlobals!()) }), {});
 }
+
+export async function transformResourcesUpdates(
```
tomeresk, line 44: You didn't actually call this in `loadPlugins`

services/src/modules/plugins/index.ts
```diff
@@ -39,3 +40,23 @@ export async function buildArgumentInjectionGlobals(): Promise<Record<string, un
...
+): Promise<Partial<ResourceGroup>> {
+  return plugins
+    .filter(p => p.transformResourcesUpdates)
+    .reduce<Partial<ResourceGroup> | Promise<Partial<ResourceGroup>>>(
+      async (res, cur) => cur.transformResourcesUpdates!(await res),
+      resourceGroup
+    );
+}
+
+export async function transformResourceGroup(resourceGroup: ResourceGroup): Promise<ResourceGroup> {
```
tomeresk, line 55: You didn't actually call this in `loadPlugins`

services/src/registry.ts
```diff
@@ -18,6 +19,7 @@ async function run() {
   await opaHelper.initializeForRegistry();
 
   logger.info('Stitch registry booting up...');
+  await loadPlugins();
```
tomeresk, line 22: Maybe we want to add a flag here to load only plugin types that are relevant to registry?

services/src/modules/plugins/index.ts
AleF83: I think that name collision is the deployer problem. Adding namespaces will add a lot of verbosity. Private plugins' variables can be edited in case of collision and public plugins' variables should be well-named.

services/src/modules/plugins/index.ts
AleF83: It shouldn't be called in `loadPlugins` it's called in `fetchAndValidate`.

services/src/modules/plugins/index.ts
AleF83: It shouldn't be called in loadPlugins it's called in fetchAndValidate.

services/src/registry.ts
AleF83: Loading plugins is just creation of object array.
The only thing we can skip in registry is `buildArgumentInjectionGlobals`.
I'm not sure it worth adding flag.

services/src/modules/plugins/index.ts
AleF83: For some reason `import fs from 'fs/promises'` doesn't work. So I've changed to `import { promises as fs } from 'fs'`

tomeresk (approved):

services/src/modules/plugins/index.ts
AleF83: Yeah, I tried to change to `lts` but I think that it's still 12.
It works with 14. But I don't want to change because we have 3 more places with the same import.
We'll change it in some upgrade dependencies PR.

PR #195, opened 2020-10-21
Capture container logs to E2E tests snapshots

AleF83:
This PR adds helper to capture container logs.
The logs may be compared with previously saved in snapshot ones.


services/tests/helpers/get-container-logs.ts
```diff
@@ -0,0 +1,79 @@
+//import * as childProcess from 'child_process';
```
tomeresk, line 1: Can probably be removed?

tomeresk (approved):

PR #197, opened 2020-10-22
Audience validation added

AleF83:


services/tests/e2e/authentication/verify-jwt.spec.ts
```diff
@@ -30,13 +36,10 @@ describe('Authentication - Verify JWT', () => {
...
     const result = await gatewayClient.request(query);
-    const captureResult = await endContainerOutputCapture();
-    const logs = captureResult
-      .split('\n')
-      .filter(line => !line.includes('    at ') && !line.includes('[deprecated]'))
-      .join('\n');
-
     expect(result).toMatchSnapshot();
+
+    const captureResult = await endContainerOutputCapture();
```
tomeresk, line 41: If the `expect` above this fails, this line would never get executed and the container output capture would not be restored to default (also true for the other test cases ending the capture after the expect)

tomeresk (approved):

services/tests/e2e/authentication/verify-jwt.spec.ts
AleF83: Not only `expect` if the request to `gateway` fails this will fails the whole test run.
I think the solution can be sending some string to `startOutputCapture` and so each logs chunk will be wrapped with its own flags. I'll add it in new PR. WDYT?

PR #201, opened 2020-10-25
Configure 404 response behaviour for @rest directive

AleF83:


AvivRubys (commented):
What do you think about just making this the default? It seems very odd someone would need to modify this, better to add less APIs if possible

tomeresk:
> What do you think about just making this the default? It seems very odd someone would need to modify this, better to add less APIs if possible

It is already the default in this PR, so it's just a matter of whether we keep the param that allows modifying it or not.
Since it's an optional param that most people will probably ignore and not send in most cases, I don't really feel strongly in either direction, both options seem fine to me

tomeresk (approved):

PR #202, opened 2020-10-26
Configure authority and JWK retriever options

AleF83:


services/src/modules/authentication/get-secret.ts
```diff
@@ -10,22 +11,43 @@ const createClient = (jwks as unknown) as (options: ClientOptions) => JwksClient
...
+  }
+}
+
+const defaultConfig: Partial<ClientOptions> = {
+  cacheMaxAge: 24 * 60 * 60 * 1000,
+};
+
+async function getJwkClient(authority: string, jwksConfig: Partial<ClientOptions>) {
+  if (!jwkClients[authority]) {
+    const jwksUri = await getJwksUri(authority);
```
tomeresk, line 31: If `jwksConfig` contains `jwksUri`, it will override the one we retrieve here from the authority (as it should).
Probably worth checking if the config contains it before retrieving it here, it will save a redundant call to the authority.

tomeresk (approved):

services/src/modules/authentication/get-secret.ts
AleF83: You're right. I've fixed

PR #205, opened 2020-11-08
Reject requests with auth headers in anonymous strategy

AleF83:


services/src/modules/authentication/anonymous-plugin.ts
```diff
@@ -0,0 +1,24 @@
...
+
+declare module 'fastify' {
+  interface FastifyRequest<
+    HttpRequest = http.IncomingMessage,
+    Query = DefaultQuery,
+    Params = DefaultParams,
+    Headers = DefaultHeaders,
+    Body = DefaultBody
+  > {
+    isAnonymousAccess(): boolean | undefined;
```
tomeresk, line 13: I don't really see a reason for the convenience function to have a possible return value of undefined for a boolean value, just change the function implementation to `return this._isAnonymousAccess ?? false;`

tomeresk (approved):

tomeresk (approved):

PR #208, opened 2020-11-16
Argument injection supports request headers

AleF83:


services/src/modules/arguments-injection/arguments-injection.spec.ts
```diff
@@ -28,6 +28,25 @@ const testCases: [string, TestCase][] = [
...
+            return;
+          },
+        },
+        exports: { resolve: () => ({}) },
+      },
+      expected: 'some-service',
+    },
+  ],
+  [
+    'From headers',
```
yoavkarako, line 49: I think the test name got mixed up

AleF83: Fixed

tomeresk (approved):

PR #210, opened 2020-11-17
Add required flag to @rest directive query params

stephaniemelton:
This PR is based on the slack discussion and recommendations [here](https://asurion-teams.slack.com/archives/CRLS8U5UM/p1605544550009400).

This is to add in a _required_ flag to the @rest query: [KeyValue], to ensure the API request is only sent if required query params were provided. With this change, we can avoid sending invalid request, and improve response times (due to not waiting on failed requests). 

- [x] All tests pass
- [x] New tests added
- [x] Documentation updated

### DIFF
--- a/services/src/modules/directives/rest/datasource.spec.ts
+++ b/services/src/modules/directives/rest/datasource.spec.ts
@@ -1,6 +1,7 @@
 import 'jest-fetch-mock';
 import { InMemoryLRUCache } from 'apollo-server-caching';
 import { RESTDirectiveDataSource } from './datasource';
+
 jest.mock('../../logger');
 
 const emptyContext = {
@@ -88,6 +89,69 @@ describe('REST directive data source', () => {
     expect(req.url).toBe('http://somewhere/');
   });
 
+  it('Sends request if required headers are included', async () => {
+    await ds.doRequest(
+      {
+        url: 'http://somewhere',
+        headers: [
+          { key: 'mykey', value: 'myvalue', required: true },
+          { key: 'owner', value: '{args.name}' },
+        ],
+      },
+      null,
+      { name: 'aviv' },
+      {} as any,
+      {} as any
+    );
+
+    expect(fetchMock).toHaveBeenCalledTimes(1);
+    const req = fetchMock.mock.calls[0][0] as Request;
+    expect(req.headers.get('mykey')).toBe('myvalue');
+    expect(req.headers.get('owner')).toBe('aviv');
+  });
+
+  it('Does not send request if required headers are empty', async () => {
+    try{
+      await ds.doRequest(
+        {
+          url: 'http://somewhere',
+          headers: [
+            { key: 'mykey', value: '', required: true },
+            { key: 'owner', value: '{args.name}' },
+          ],
+        },
+        null,
+        { name: 'aviv' },
+        {} as any,
+        {} as any
+      );
+    }catch(err) {
+      expect(err.toString()).toBe("mykey header is required");
+    }
+    expect(fetchMock).toHaveBeenCalledTimes(0);
+  });
+
+  it('Sends request when header is missing the value and required is not set (default false)', async () => {
+    await ds.doRequest(
+      {
+        url: 'http://somewhere',
+        headers: [
+          { key: 'mykey', value: '' },
+          { key: 'owner', value: '{args.name}' },
+        ],
+      },
+      null,
+      { name: 'aviv' },
+      {} as any,
+      {} as any
+    );
+
+    expect(fetchMock).toHaveBeenCalledTimes(1);
+    const req = fetchMock.mock.calls[0][0] as Request;
+    expect(req.headers.get('mykey')).toBe('');
+    expect(req.headers.get('owner')).toBe('aviv');
+  });
+
   describe('Supported http methods dispatch with the correct http method', () => {
     const methods = ['GET', 'DELETE', 'POST', 'PUT', 'PATCH'];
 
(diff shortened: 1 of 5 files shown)


services/src/modules/directives/rest/datasource.ts
```diff
@@ -86,10 +89,16 @@ export class RESTDirectiveDataSource extends RESTDataSource<RequestContext> {
...
     info: GraphQLResolveInfo
   ) {
-    if (!kvs || kvs.length == 0) return;
+    let validQueryParams = true;
+
+    if (!kvs || kvs.length == 0) return validQueryParams;
 
     for (const kv of kvs) {
       const value = inject(kv.value, parent, args, this.context, info);
+      if (!value && kv.required) {
```
AleF83, line 98: This doesn't work for `false`.
```suggestion
      if (value === undefined && kv.required) {
```

services/src/modules/directives/rest/datasource.ts
```diff
@@ -86,10 +89,16 @@ export class RESTDirectiveDataSource extends RESTDataSource<RequestContext> {
...
+        validQueryParams = false;
+        break;
+      }
       if (!value) {
```
AleF83, line 102: We already have bug here :-)
```suggestion
      if (value === undefined) {
```

services/src/modules/directives/rest/datasource.ts
```diff
@@ -86,10 +89,16 @@ export class RESTDirectiveDataSource extends RESTDataSource<RequestContext> {
...
   ) {
-    if (!kvs || kvs.length == 0) return;
+    let validQueryParams = true;
+
+    if (!kvs || kvs.length == 0) return validQueryParams;
 
     for (const kv of kvs) {
       const value = inject(kv.value, parent, args, this.context, info);
+      if (!value && kv.required) {
+        validQueryParams = false;
```
AleF83, line 99: If required query param is missing the function should throw `GraphQLError`.
The field in response will be `null` and there will be error in `errors` filed.
```suggestion
        throw new GraphQLError(`${kv.key} query parameter is required`);
```

services/src/modules/directives/rest/types.ts
```diff
@@ -1,6 +1,7 @@
 export interface KeyValue {
   key: string;
   value: string;
+  required?: boolean;
```
AleF83, line 4: This interface is in use for request headers too. So we need to implement the same functionality in `parseHeaders` function.

AleF83 (changes requested):

services/src/modules/directives/rest/types.ts
tomeresk: Makes sense to also support the same logic for optionally requiring headers

services/src/modules/directives/rest/datasource.spec.ts
```diff
@@ -131,4 +131,39 @@ describe('REST directive data source', () => {
     expect(req.headers.get('Content-Type')).toBe('application/json');
     expect(await req.text()).toBe('{"name":"aviv"}');
   });
+
+  it('Sends request if required query parameters are included', async () => {
```
tomeresk, line 135: Probably worth adding 1 more case that tests for backward compatibility (the default false value for required) - a query parameter missing the value with the required field not specified at all

services/src/modules/directives/rest/datasource.ts
stephaniemelton: There is a valid scenario which causes us to not want to add this throw error here. Which is the use case I am trying to solve for with this PR. Using the rest directive on a field instead of a query, means that there are some cases when the source might not have a value that is being used as input to the rest directive. For example, I am using @rest to resolve a billing field on an agreement (parent type). Some agreements do not have billing ids associated with it (ie, client is handling billing vs us). In those cases the source record will not return billing ids which is what @rest uses as input for resolving this field. That doesn't mean its wrong, it just doesn't have enough source data to resolve the field. What are your thoughts around this scenario?

services/src/modules/directives/rest/datasource.ts
stephaniemelton: value in the KeyValue type is a string, so it wouldn't let me set it as undefined or a boolean in the unit tests. When I changed this to value === undefined, it was still sending requests, and not entering into this failure statement.

services/src/modules/directives/rest/datasource.ts
stephaniemelton: Same comment as above. I don't think this was an bug.

services/src/modules/directives/rest/datasource.ts
stephaniemelton: Additional note in general on this - I added the throw statement, however, the initial use case I was trying to account for still ends up sending the request since I am using argument injection to resolve the input query parameter. If there is no match, its returning whatever the input was. The api request sends with argument injection as the value for the query parameter. Any thoughts on how to get around this?

AleF83: Following our discussion in Slack.
1) The first point will be resolved by adding optional parameter `condition` to `@localResolver` directive.
2) The second point can be resolved by changing the `value` expression to return `null` in the right cases. For example, instead of `"{source.foo.bar}"` when `foo` may be null you can set `"{source.foo && source.foo.bar}"`.
In this way you'll avoid an exception in argument evaluation and the return value will be `null`

services/src/modules/directives/rest/datasource.ts
AleF83: The `value` is result of argument evaluation. It can be anything. See the code 4 lines below.
https://github.com/Soluto/stitch/blob/fb8e62878d0ed4c89bc3ffe631aa965ce9a7d12e/services/src/modules/directives/rest/datasource.ts#L96

In `parseHeaders` function there is casting to `string`. https://github.com/Soluto/stitch/blob/fb8e62878d0ed4c89bc3ffe631aa965ce9a7d12e/services/src/modules/directives/rest/datasource.ts#L75
I think it can work for headers because despite the query params headers should be unique.

AleF83 (approved):

tomeresk (approved):

PR #217, opened 2020-11-29
CLI: Exclude resource types

AleF83:


cli/oclif.manifest.json
```diff
@@ -57,7 +57,8 @@
           "type": "option",
           "description": "Custom authorization header",
           "required": false
-        }
+        },
+        "exclude": { "name": "exclude", "type": "option", "description": "Resource types to skip", "required": false }
```
tomeresk, line 61: I think `exclude` is too general, maybe `exclude-resource-types` or `ignore-resource-types`?

Also description should specify how to send multiple types, e.g. `A comma separated list of resource types to skip/ignore` (skip or ignore based on chosen name)

There is also an option that allows setting a flag multiple times (`multiple: true` https://oclif.io/docs/flags), not sure how the output is presented based on the docs but worth checking

cli/tests/apply-resources.spec.ts
```diff
@@ -12,10 +12,17 @@ describe('Apply resources', () => {
       })
     )
     .stdout()
-    .command(['apply:resources', '--dry-run', '--registry-url=http://registry/graphql', 'tests/resources'])
+    .command([
+      'apply:resources',
+      '--dry-run',
+      '--registry-url=http://registry/graphql',
+      '--exclude=upstreams',
```
tomeresk, line 19: Should also add a test for skipping multiple resource types

tomeresk (changes requested):

cli/oclif.manifest.json
AleF83: Renamed to `skip-resource-types`.
I don't think that this is the case of `multiple` flag. It's more suitable for something like `-f` in `kubeclt` command.

tomeresk (approved):
Still missing the mention of the comma separated list in the description

PR #218, opened 2020-12-02
CLI: Schema resource type support multiple files

AleF83:
`Schema` resources can have inline `schema` property or `schemaFiles` one:

```yaml
metadata:
  namespace: <namespace>
  name: <name>
schemaFiles:
  - ./path/to/files/**/*.gql
  - ./one/more/file.gql
```

The content of all these files is concatenated to one sdl block and is handled as one `schema` property.

The test framework of CLI was changed from `jest` to `mocha` due to some issues in `@oclif/test` library. It's based on `mocha` and its `jest` support has leaks that causes process hang up.


cli/src/commands/apply/resources.ts
```diff
@@ -90,7 +91,11 @@ Uploaded successfully!
         // Client-side validation can be added later on to not require a server hop
         switch (kind) {
           case 'Schema':
-            rg.schemas!.push(resource);
+            if (resource.schema) {
```
AvivRubys, line 94: Throw error if it has neither?

cli/src/utils/build-stitched-schema.ts
```diff
@@ -0,0 +1,18 @@
+import { promises as fs } from 'fs';
+import * as globby from 'globby';
+import { ResourceMetadataInput, SchemaInput } from '../client';
+
+interface SchemaFilesInput {
+  metadata: ResourceMetadataInput;
+  schemaFiles: string[];
+}
+
+export default async function ({ metadata, schemaFiles }: SchemaFilesInput): Promise<SchemaInput> {
```
AvivRubys, line 10: I'd change the name of this file, I thought it was doing some graphql federation/schema stitching work
Maybe `readSchemaFiles` or something like that? It's just concatting some files together

cli/src/commands/apply/resources.ts
AleF83: We don't throw error on other resources.. 
I think the error will be thrown either by `graphql-request` or by Stitch Registry.

cli/src/utils/build-stitched-schema.ts
AleF83: Fixed

cli/src/commands/apply/resources.ts
```diff
@@ -3,6 +3,7 @@ import * as path from 'path';
 import { Command, flags } from '@oclif/command';
 import { safeLoadAll } from 'js-yaml';
 import { ResourceGroupInput, uploadResourceGroup } from '../../client';
+import buildStitchedSchema from '../../utils/read-schema-files';
```
tomeresk, line 6: Should probably reflect the new file name here

```suggestion
import readSchemaFiles from '../../utils/read-schema-files';
```

I also think `concatenateSchemaFiles` would be a more descriptive name for the file

tomeresk (approved):

PR #222, opened 2020-12-06
Registry Metrics endpoint with default metrics

AleF83:


services/tests/e2e/basic/metrics.spec.ts
```diff
@@ -74,4 +78,12 @@ describe('Metrics', () => {
       )
     ).toBeTruthy();
   });
+
+  test('Registry metrics', async () => {
+    const response = await fetch(registryMetricsEndpoint);
+    expect(response.ok).toBeTruthy();
+    const body = await response.text();
+    const metrics = body.split('\n');
+    expect(metrics.length).toBeGreaterThan(0);
```
tomeresk, line 87: An elaborate error message would also pass this test technically, might be better to look for some very basic metric that we are pretty sure will stay there

tomeresk (approved):

PR #224, opened 2020-12-07
Registry resolvers refactor

AleF83:
Small code simplification


tomeresk (approved):
LGTM, the function in the `update-resource-group` file is a bit complicated after removing the `fetchAndValidate` function, but I don't really have a better idea since most of the variables are required in many places in the function (which also caused the weird interactions with the old `fetchAndValidate` function)

PR #225, opened 2020-12-07
Plugins: add configure method

AleF83:


docs/plugins.md
```diff
@@ -31,6 +33,12 @@ export interface StitchPlugin {
 
 ## API Reference
 
+### configure
+
+Optional method that's called on plugin initialization. Can receive some configuration from the Stitch.
```
tomeresk, line 38:
```suggestion
Optional method that's called on plugin initialization. Stitch will send it the configuration relevant for this plugin, based on the values defined in the `PLUGINS_CONFIGURATION` environment variable.

For example, with this env var value:
`PLUGINS_CONFIGURATION={ "awesome-plugin.js": { "arg1": "val1" } }`

The `options` param for the `configure` function will be `{ arg: 'val1' }`
```

tomeresk (approved):
LGTM, docs should be clearer though

PR #234, opened 2020-12-12
Refactor rest directive for future upstream resource change

AleF83:


services/src/modules/directives/rest/datasource.ts
```diff
@@ -73,36 +43,70 @@ export class RESTDirectiveDataSource extends RESTDataSource<RequestContext> {
...
+      throw new Error('Set either "body" or "bodyArg" argument but not both');
+    }
+    const body = bodyObj
+      ? inject(bodyObj, fieldResolverParams)
+      : fieldResolverParams.args[params.bodyArg || 'input'];
+    if (body) {
+      requestInit.body = JSON.stringify(body);
+      if (!requestInit.headers) {
+        requestInit.headers = new Headers();
+      }
```
tomeresk, line 61: This seems out of place here, just make `setHeaders` always set this, even if there are no headers to fill. It currently already always initializes (`const headers = new Headers();` in line 33) so just set this as well in any case instead of only at the end

tomeresk (approved):

PR #239, opened 2020-12-15
Gql directive: move introspect query to registry

AleF83:


services/src/modules/directives/gql/directive.ts
```diff
@@ -0,0 +1,73 @@
...
+import { RequestContext } from '../../context';
+import { getAuthHeaders } from '../../upstreams/authentication';
+import { RemoteSchema } from './introspection';
+
+export class GqlDirective extends SchemaDirectiveVisitor {
+  visitFieldDefinition(field: GraphQLField<unknown, RequestContext>) {
+    const { url, fieldName, arguments: gqlArgs, operationType: operationTypeParam, timeoutMs } = this.args;
+    const operationType = operationTypeParam?.toLowerCase() ?? 'query';
+
+    if (!this.context?.resourceGroup) return;
```
tomeresk, line 19: We probably want an actual error thrown here?

services/src/modules/plugins/transform-resource-group.spec.ts
```diff
@@ -20,6 +20,7 @@ const resourceGroup: ResourceGroup = {
   upstreams: [],
   upstreamClientCredentials: [],
   policies: [],
+  remoteSchemas: [],
```
tomeresk, line 23: Should we initialize this as empty here? It is a "calculated" field like policy attachment so we should be able to handle cases where it's missing (e.g. if there are no gql directives at all in the schema)

services/src/modules/validation.spec.ts
```diff
@@ -15,6 +15,7 @@ describe('validateMetadataCharacters - Validation that namespaces and names cont
       upstreams: [],
       upstreamClientCredentials: [],
       policies: [{ metadata: { namespace: 'other-ns', name: 'valid-chars_only' }, type: PolicyType.opa, code: 'code' }],
+      remoteSchemas: [],
```
tomeresk, line 18: Also here, we probably shouldn't have to initialize it as empty if it's not a permanent part of the resource group

services/tests/e2e/docker-compose.yml
```diff
@@ -74,8 +74,8 @@ services:
     command: >
       -c "
       mkdir -p /var/minio/${S3_RESOURCE_BUCKET_NAME} &&
-      echo '{\"schemas\":[],\"upstreams\": [],\"upstreamClientCredentials\": [],\"policies\": []}' > /var/minio/${S3_RESOURCE_BUCKET_NAME}/${S3_RESOURCE_OBJECT_KEY} &&
-      echo '{\"schemas\":[],\"upstreams\": [],\"upstreamClientCredentials\": [],\"policies\": []}' > /var/minio/${S3_RESOURCE_BUCKET_NAME}/${S3_REGISTRY_RESOURCE_OBJECT_KEY} &&
+      echo '{\"schemas\":[],\"upstreams\": [],\"upstreamClientCredentials\": [],\"policies\": [], \"remoteSchemas\": []}' > /var/minio/${S3_RESOURCE_BUCKET_NAME}/${S3_RESOURCE_OBJECT_KEY} &&
```
tomeresk, line 77: and here

tomeresk (changes requested):

services/src/modules/directives/gql/directive.ts
```diff
@@ -0,0 +1,71 @@
...
+import { getAuthHeaders } from '../../upstreams/authentication';
+import { RemoteSchema } from './introspection';
+
+export class GqlDirective extends SchemaDirectiveVisitor {
+  visitFieldDefinition(field: GraphQLField<unknown, RequestContext>) {
+    const { url, fieldName, arguments: gqlArgs, operationType: operationTypeParam, timeoutMs } = this.args;
+    const operationType = operationTypeParam?.toLowerCase() ?? 'query';
+
+    const remoteSchemas: RemoteSchema[] = this.context.resourceGroup.remoteSchemas;
+    const remoteSchemaResource = remoteSchemas.find(rs => rs.url === url)?.schema;
```
tomeresk, line 20:
```suggestion
    const remoteSchemaResource = remoteSchemas?.find(rs => rs.url === url)?.schema;
```

tomeresk (approved):

PR #240, opened 2020-12-20
Fix for moving introspection query of gql directive to registry

AleF83:


services/src/modules/directives/gql/introspection.ts
```diff
@@ -15,39 +15,50 @@ export interface RemoteSchema {
...
+        if (!gqlDirectives) continue;
+
+        for (const gqlDirective of gqlDirectives) {
+          if (!gqlDirective.arguments) continue;
+
+          const urlArgument = gqlDirective.arguments!.find(a => a.name.value === 'url');
+          if (!urlArgument) continue;
+          const url = (urlArgument!.value as StringValueNode).value;
+
+          if (!knownRemoteGqlServers.has(url)) {
```
tomeresk, line 41: Doesn't this mean we never update the existing remote schemas? why not update them?

tomeresk (approved):

services/src/modules/directives/gql/introspection.ts
AleF83: I think we should update the remote schemas from time to time. But I'm not sure that the trigger to this update should be update of other not related resource in resource group.

PR #246, opened 2020-12-29
Delete resources mutations

AleF83:


tomeresk (approved):
Missing docs here, though I see we don't really have registry docs in general that this can be added to so I'm approving it (we need to add docs for registry endpoints in general in another task)

PR #248, opened 2020-12-31
Policy definition args - allow setting a default value for an arg and…

tomeresk:
… allow making it optional


services/src/modules/resource-repository/types.ts
```diff
@@ -62,11 +63,17 @@ export interface UpstreamClientCredentials extends Resource {
...
-  args?: PolicyArgsObject;
+  args?: PolicyArgsDefinitions;
   query?: PolicyQuery;
   shouldOverrideBasePolicy?: boolean;
 }
 
+export interface PolicyArgDefinition {
+  type: string;
+  default?: string;
+  optional?: boolean;
```
AleF83, line 74: I think we can make this field internal. It can be set in registry by calling `parseType` and `isNotNullType` functions from `graphql`. So we can validate that user provided valid type and decide if the argument is required or optional according to the type itself

services/src/modules/directives/policy/policy-executor.ts
```diff
@@ -112,22 +112,30 @@ export default class PolicyExecutor {
...
-        const { source, gqlArgs: args, requestContext: context, info } = ctx;
-        policyArgValue = inject(policyArgValue, { source, args, context, info });
-      }
-
-      policyArgs[policyArgName] = policyArgValue;
-      return policyArgs;
-    }, {});
+    return Object.entries(supportedPolicyArgs).reduce<PolicyArgsObject>(
+      (policyArgs, [policyArgName, { default: defaultArg, optional = false }]) => {
+        let policyArgValue = ctx.policy?.args?.[policyArgName] ?? defaultArg;
```
AleF83, line 117: I think we should distinguish between `null` and `undefined` values. `null` should be assumed as valid argument value and we shouldn't replace it with the default

services/src/modules/registry-schema/helpers/index.ts
```diff
@@ -0,0 +1,3 @@
+import proceedPolicy from './policy-helper';
```
tomeresk, line 1: Not a fan of the`proceedPolicy` name here, the function does a very specific function but the name is quite generic. How about `markOptionalPolicyArgs`?

Regarding the file, it is named `policy-helper` but exports some specific utility function as default. We should either rename the file specifically after that function, or change the export to be non-default

AleF83 (approved):

PR #255, opened 2021-01-09
Bug fix: Enable to use FS resource repo in registry

AleF83:


services/src/modules/resource-repository/get-resource-repository.ts
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  switch (repositories.length) {
+    case 0:
+      logger.fatal('Must enable at least one resource repository');
+      throw new Error('Must enable at least one resource repository');
+    case 1:
+      return repositories[0];
+    default:
+      return new CompositeResourceRepository(repositories);
```
tomeresk, line 29: This makes no sense for registry since you can't do any updates with the composite repository (I guess that was the reason for the original separation?)
Anyway it probably is better to have this function in one clear place, and we already have the `isRegistry` data here, so I think this should be good:

```suggestion
    default:
      if (options?.isRegistry) {
        logger.fatal('Registry cannot have more than one resource repository');
        throw new Error('Registry cannot have more than one resource repository');
      }
      return new CompositeResourceRepository(repositories);
```

I would probably make a small function for the `logger.fatal` and throwing error since we have to repeat a message twice now (both in this case and in the case of 0 resource repositories)

tomeresk (changes requested):

services/src/modules/resource-repository/get-resource-repository.ts
AleF83: These are two different messages. I don't think there is code reuse here.

tomeresk: I mean the fact that you send the same message twice (in both places) to both the `logger.fatal` and the `new Error`

tomeresk (approved):

PR #257, opened 2021-01-09
Blackbox tests

AleF83:


services/tests/blackbox/upstreams/upstream-for-gql.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+  const jwtIssuer = 'https://oidc-provider';
+
+  let app: FastifyInstance;
+  let dispose: () => Promise<void>;
+
+  let remoteServerScope: nock.Scope;
+
+  let token: string;
+
+  beforeAll(async () => {
```
tomeresk, line 25: This `beforeAll` is a bit much, I would take some of it out to functions

tomeresk (approved):

PR #261, opened 2021-01-14
More logs for registry

AleF83:


services/src/modules/registry-schema/resolvers/delete-resources.ts
```diff
@@ -19,40 +19,53 @@ export default async function (deletions: ResourceGroupMetadataInput, context: R
...
     } finally {
+      logger.trace('Removing policy attachments temporary files...');
       await policyAttachments.cleanup();
     }
 
+    const summary = Object.fromEntries(
+      Object.entries(deletions).map(([k, v]) => [k, v ? (Array.isArray(v) ? v.length : 1) : 0])
+    );
+    logger.info(summary, `Resources were deleted`);
+
```
yoavkarako, line 68: Why remove from the `try` clause?

services/src/modules/logger.ts
```diff
@@ -11,7 +11,7 @@ const devLoggerConfig: pino.LoggerOptions = {
   level: 'trace',
   prettyPrint: {
     colorize: true,
-    translateTime: 'HH:MM:ss',
+    translateTime: 'HH:MM:ss.l',
```
yoavkarako, line 14: Which formatting engine is being used here where `l` correlates to ms?

AleF83: `pino` logger uses https://github.com/felixge/node-dateformat

services/src/modules/registry-schema/resolvers/delete-resources.ts
AleF83: This logs end of operation. So for exact time measurement it should be done after `finally` clause.

yoavkarako: Do you want to log even if there's an exception and the process fails?

AleF83: If exception will be thrown this code won't run.

yoavkarako (approved):

tomeresk (approved):

PR #262, opened 2021-01-15
CLI refactor and logs

AleF83:


cli/src/utils/list-resource-files.ts
```diff
@@ -0,0 +1,23 @@
...
+    for (const file of files) {
+      await walk(fileList, join(path, file), extensions);
+    }
+    return;
+  }
+  const isResource = extensions.some(ext => path.endsWith(`.${ext}`));
+  if (isResource) {
+    fileList.push(path);
+  }
+}
```
tomeresk, line 23:
```suggestion
export default async function walk(path: string, extensions = ['yaml', 'yml'], fileList: string[] = []) {
  const stat = await fs.stat(path);
  if (stat.isDirectory()) {
    const files = await fs.readdir(path);
    for (const file of files) {
      await walk(join(path, file), extensions, fileList);
    }
    return;
  }
  const isResource = extensions.some(ext => path.endsWith(`.${ext}`));
  if (isResource) {
    fileList.push(path);
  }
}
```

tomeresk (approved):

cli/src/utils/list-resource-files.ts
AleF83: According to your suggestion array should be created in function that calls to this one. I think it's not so clear why `listResourceFiles` function gets empty array to fill in other file without seeing the code of the function.

I think no need to add default value to the `fileList` parameter.

tomeresk: @AleF83 no that was not the suggestion, with my suggestion calling it would be exactly like calling the `walk` function in your suggestion, the calling function does not need to send the array and it can send only the `path` parameter. You can call it `_fileList` if you want to make that clearer.

Callling function would call my suggestion like this: `const resultList = await walk(somePath)`.

If you are concerned about "exposing" the fileList parameter (though I don't see it as an issue, it has a default value and should just not be sent unless you know what you are doing, it's a pretty standard practice in recursive functions), you can also keep the skeleton function that calls it and just call it without all the initialization:

```suggestion
export default function (path: string, extensions = ['yaml', 'yml']) {
  return walk(path, extensions);
}

export async function walk(path: string, extensions = ['yaml', 'yml'], fileList: string[] = []) {
  const stat = await fs.stat(path);
  if (stat.isDirectory()) {
    const files = await fs.readdir(path);
    for (const file of files) {
      await walk(join(path, file), extensions, fileList);
    }
    return;
  }
  const isResource = extensions.some(ext => path.endsWith(`.${ext}`));
  if (isResource) {
    fileList.push(path);
  }
}
```

PR #265, opened 2021-01-21
Support custom logger configuration

AleF83:
Now Stitch logger can be configured via environment variable.
The configuration is `pino` logger configuration. Properties of config object that are functions aren't supported.

### DIFF
--- a/docs/logging.md
+++ b/docs/logging.md
@@ -1,0 +1,18 @@
+# Logging
+
+Stitch uses [`pino`](https://github.com/pinojs/pino) logger. The logger default configuration can be finded [here](../services/src/modules/logger.ts).
+
+Custom configuration can be provided using `LOGGER_CONFIGURATION` environment variable.
+
+> Note: environment variable value should be JSON-serialized object of [`pino.LoggerOptions`](https://getpino.io/#/docs/api?id=options-object) type. Properties of `function` type aren't supported.
+
+Example:
+
+```yaml
+LOGGER_CONFIGURATION: |
+  {
+    "redact": [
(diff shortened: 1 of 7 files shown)


docs/logging.md
```diff
@@ -0,0 +1,18 @@
+# Logging
+
+Stitch uses [`pino`](https://github.com/pinojs/pino) logger. The logger default configuration can be finded [here](../services/src/modules/logger.ts).
```
tomeresk, line 3:
```suggestion
Stitch uses [`pino`](https://github.com/pinojs/pino) logger. The logger's default configuration can be found [here](../services/src/modules/logger.ts).
```

services/src/modules/config.ts
```diff
@@ -1,5 +1,6 @@
 import * as path from 'path';
 import * as envVar from 'env-var';
+import * as pino from 'pino';
```
tomeresk, line 3:
```suggestion
import { LoggerOptions } from 'pino';
```

services/src/modules/config.ts
```diff
@@ -55,3 +56,5 @@ export const knownApolloDirectives = envVarExt
   .get('KNOWN_APOLLO_DIRECTIVES')
   .default(defaultKnownApolloDirectives)
   .asSet();
+
+export const loggerConfiguration = envVar.get('LOGGER_CONFIGURATION').asJsonObject() as pino.LoggerOptions;
```
tomeresk, line 60: Probably want to default that to an empty object somewhere here

tomeresk (approved):

PR #271, opened 2021-02-08
@enumResolver directive

AleF83:
`@enumResolver` and `@enumValue` directives added. This directive allows to define graphql enums like typescript enums:

```typescript
enum Foo {
  Bar = 'bar',
}
```

Usage example:

```graphql
enum Foo @enumResolver {
   Bar @enumValue(value: "bar")
}
```

### DIFF
--- a/services/src/modules/build-federated-schema.ts
+++ b/services/src/modules/build-federated-schema.ts
@@ -5,6 +5,8 @@
 import {
   DocumentNode,
   visit,
+  EnumValueDefinitionNode,
+  EnumTypeDefinitionNode,
   FieldDefinitionNode,
   ASTNode,
   InterfaceTypeDefinitionNode,
@@ -22,7 +24,6 @@ import { composeAndValidate } from '@apollo/federation';
 import { ApolloError } from 'apollo-server-core';
 import createTypeResolvers from './implicit-type-resolver';
 import { knownApolloDirectives } from './config';
-
 interface DirectiveVisitors {
   [directiveName: string]: typeof SchemaDirectiveVisitor;
 }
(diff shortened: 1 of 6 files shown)


services/src/modules/build-federated-schema.ts
```diff
@@ -134,12 +154,42 @@ function addDirectivesToTypeDefs(
...
+    },
+
+    EnumValueDefinition(node, _key, _parent, _path, ancestors) {
+      const enumNode = ancestors[ancestors.length - 1] as ASTNode;
+      if (!isEnumTypeDefinitionNode(enumNode)) {
+        throw new Error(`Expected EnumTypeDefinitionNode, found ${enumNode.kind}`);
+      }
+
+      const enumValDirectives =
+        enumValueDirectives[enumNode.name.value] && enumValueDirectives[enumNode.name.value][node.name.value];
```
tomeresk, line 176:
```suggestion
      const enumValDirectives = enumValueDirectives[enumNode.name.value]?.[node.name.value];
```

services/src/modules/build-federated-schema.ts
```diff
@@ -249,3 +350,46 @@ function mergeFieldDefinitionDirectivesUsages(dus: DirectivesUsagesByObjectAndFi
...
+  for (const du of dus) {
+    for (const objectName in du) {
+      const usages = du[objectName];
+
+      if (!(objectName in result)) {
+        result[objectName] = {};
+      }
+
+      result[objectName] = usages;
+    }
```
tomeresk, line 366:
```suggestion
    for (const [objectName, usages] of Object.entries(du)) {
      if (!(objectName in result)) result[objectName] = {};

      result[objectName] = usages;
    }
```

services/src/modules/build-federated-schema.ts
```diff
@@ -249,3 +350,46 @@ function mergeFieldDefinitionDirectivesUsages(dus: DirectivesUsagesByObjectAndFi
...
+  return result;
+}
+
+function mergeEnumValueDirectivesUsages(dus: DirectivesUsagesByEnumNameAndValue[]) {
+  const result: DirectivesUsagesByEnumNameAndValue = {};
+
+  for (const du of dus) {
+    for (const objectName in du) {
+      const fields = du[objectName];
+      for (const fieldName in fields) {
```
tomeresk, line 378: Can also use `Object.entries` for the two `for...in` loops here

tomeresk (approved):

PR #274, opened 2021-02-15
add resolver info param to injection data

yoavkarako:


AleF83 (changes requested):
LGTM,
Please update docs: https://github.com/Soluto/stitch/blob/master/docs/arguments_injection.md

AleF83 (approved):

PR #275, opened 2021-02-17
Refresh remote schema mutation

AleF83:
### Why
On uploading schema resources with gql directive Stitch registry makes introspection query and stores the result in resource group.
Remote service schema may change without any changes in Stitch resources.

## What
We need to refresh remote schema on deployment of new version of remote graphql service.

## How
Add `refreshRemoteSchema` mutation to Stitch registry and `refresh:remote-schema` command to stitch-cli. The cli command should be part of the remote service CI/CD pipeline.

### DIFF
--- a/cli/src/client/index.ts
+++ b/cli/src/client/index.ts
@@ -109,7 +109,6 @@ export async function refreshRemoteSchema(
   url: string,
   options: {
     registryUrl: string;
-    dryRun?: boolean;
     authorizationHeader?: string;
   },
   clientOptions: Partial<RequestInit>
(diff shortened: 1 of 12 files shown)


cli/src/client/index.ts
```diff
@@ -97,4 +105,20 @@ export async function uploadBasePolicy(
   return registryClient.request<{ result: { success: boolean } }>(query, { basePolicy });
 }
 
+export async function refreshRemoteSchema(
+  url: string,
+  options: {
+    registryUrl: string;
+    dryRun?: boolean;
```
thewizarodofoz, line 112: didn't see this in the command flags

thewizarodofoz (approved):

tomeresk (approved):

PR #296, opened 2021-03-10
export plugin data to injection

yoavkarako:


services/src/modules/arguments-injection/arguments-injection.ts
```diff
@@ -25,14 +27,15 @@ function injectTemplate<T = unknown>(
     exports: info && context && getExportsProxy(context.exports, info?.parentType, source as Record<string, unknown>),
     vars: info?.variableValues,
     info,
+    pluginData: context?.resourceGroup?.pluginData,
```
AleF83, line 30: `pluginData` is a bit long.
```suggestion
    plugins: context?.resourceGroup?.pluginData,
```

services/tests/e2e/basic/plugins.schema.ts
```diff
@@ -20,6 +20,7 @@ export const schema: Schema = {
         @policy(namespace: "plugins", name: "whole-rg-policy")
       pl_new_foo: Point! @localResolver(value: { x: 1, y: 2 })
       pl_new_bar: String! @localResolver(value: "NEW BAR") @reverse
+      pl_new_data: String! @localResolver(value: "{pluginData.foo}")
```
AleF83, line 23:
```suggestion
      pl_new_data: String! @localResolver(value: "{plugins.foo}")
```

services/tests/e2e/config/plugins/transform-resource-group.js
```diff
@@ -24,6 +24,8 @@ module.exports = {
       newRg.policies.push(wholeRgPolicy);
     }
 
+    newRg.pluginData = { foo: 'bar' };
```
AleF83, line 27: `pluginData` should be initialized to an empty object before applying plugins on resource group. So plugins will add some properties to the object rather then replace/create it.

AleF83 (changes requested):
Run tests with `-u` flag to update snapshots. Check that the snapshots were changed as expected

yoavkarako:
> Run tests with `-u` flag to update snapshots. Check that the snapshots were changed as expected

Can you do that? It won't run on my machine. I'll need to set it up the wsl...

AleF83:
I've rebased a bit the branch and squashed the commits.
Just add some documentation [here](https://github.com/Soluto/stitch/blob/master/docs/arguments_injection.md) and [here](https://github.com/Soluto/stitch/blob/master/docs/plugins.md#transformresourcegroup)

AleF83 (approved):

PR #297, opened 2021-03-11
Logging more data for development

AleF83:


services/src/modules/authentication/strategies/jwt.ts
```diff
@@ -30,9 +32,9 @@ export default async function (request: fastify.FastifyRequest): Promise<void> {
     try {
       // Verify JWT signing, expiration and more
       await request.jwtVerify();
-      logger.trace({ issuer }, 'JWT verified');
+      reqLogger.trace('JWT verified');
     } catch (err) {
-      logger.debug({ err, issuer }, 'Failed to verify request JWT');
+      reqLogger.debug({ err }, 'Failed to verify request JWT');
```
yoavkarako, line 37: Where can we observe debug level logs? Will we have some indication if someone targets the service?

tomeresk (approved):

services/src/modules/authentication/strategies/jwt.ts
AleF83: I think that the right solution is to add metrics and monitor it. If there is suspicion of the case the log level should be lowered and the specific case will be investigated.

PR #299, opened 2021-03-14
Authentication - JWT strategy - multiple audiences in config

AleF83:


services/src/modules/authentication/strategies/jwt.ts
```diff
@@ -18,10 +19,15 @@ export default async function (request: fastify.FastifyRequest): Promise<void> {
...
+    const jwtAudiences = Array.isArray(decodedJWT.payload.aud)
       ? decodedJWT.payload.aud.map(a => String(a))
       : Array.of(String(decodedJWT.payload.aud));
-    if (!audience.includes(issuerConfig.audience)) {
+    const configAudiences = Array.isArray(issuerConfig.audience)
+      ? issuerConfig.audience
+      : Array.of(issuerConfig.audience);
+
+    const intersection = _.intersection(configAudiences, jwtAudiences);
+    if (!intersection || intersection.length === 0) {
```
thewizarodofoz, line 30: `_.intersection` will always return an array, we can drop `!intersection`

services/src/modules/authentication/strategies/jwt.ts
```diff
@@ -18,10 +19,15 @@ export default async function (request: fastify.FastifyRequest): Promise<void> {
 
   // Verify audience
   if (issuerConfig.audience) {
-    const audience = Array.isArray(decodedJWT.payload.aud)
+    const jwtAudiences = Array.isArray(decodedJWT.payload.aud)
       ? decodedJWT.payload.aud.map(a => String(a))
       : Array.of(String(decodedJWT.payload.aud));
-    if (!audience.includes(issuerConfig.audience)) {
+    const configAudiences = Array.isArray(issuerConfig.audience)
```
thewizarodofoz, line 25: Why no something simpler like `[].concat(issuerConfig.audience)`?

services/tests/e2e/docker-compose.yml
```diff
@@ -27,7 +27,7 @@ services:
             },
             "http://localhost:8070": {
               "authority": "http://authority-without-discrovery-endpoint",
-              "audience": "Stitch Gateway",
+              "audience": ["Stitch Gateway", "stitch gateway"],
```
thewizarodofoz, line 30: looks like now you are only testing multiple audience but not single audience

AleF83: There is another issues above on port 8060..

services/src/modules/authentication/strategies/jwt.ts
AleF83: tried now.. `Type 'string' is not assignable to type 'ConcatArray<never>'`

AleF83: This works but it's ugly `([] as string[]).concat(issuerConfig.audience)`

thewizarodofoz (approved):

PR #315, opened 2021-04-08
Api key auth strategy

AleF83:


services/src/modules/authentication/strategies/api-key.ts
```diff
@@ -0,0 +1,44 @@
...
+  const headerName = config.header;
+  if (headerName) {
+    apiKey = request.headers[headerName] as string;
+  }
+  if (!apiKey) {
+    const queryParamName = config.queryParam;
+    if (queryParamName) {
+      apiKey = request.query[queryParamName] as string;
+    }
+  }
```
tomeresk, line 20:
```suggestion
  const apiKey = (request.headers[config.header] || request.query[config.queryParam]) as string
```

services/src/modules/authentication/strategies/api-key.ts
```diff
@@ -0,0 +1,44 @@
...
+  if (!apiKey) {
+    throw new Error('Unauthorized');
+  }
```
tomeresk, line 23:
```suggestion
  if (!apiKey) throw new Error('Unauthorized');
```

services/src/modules/authentication/strategies/api-key.ts
```diff
@@ -0,0 +1,44 @@
...
+  reqLogger.trace('Api key found in request');
+
+  const apiKeyConfig = apiKeys[apiKey];
+  if (!apiKeyConfig) {
+    reqLogger.debug('Unexpected api key');
+    throw new Error('Unauthorized');
+  }
+  reqLogger.trace({ name: apiKeyConfig.name }, 'Api key found in configuration');
+
+  if (apiKeyConfig.enabled === false) {
```
tomeresk, line 34:
```suggestion
  if (!apiKeyConfig.enabled) {
```

services/src/modules/authentication/strategies/api-key.ts
```diff
@@ -0,0 +1,44 @@
...
+    reqLogger.debug('The api key is disabled');
+    throw new Error('Unauthorized');
+  }
+
+  const anonymousPaths = config.authenticatedPaths;
```
tomeresk, line 39: I don't understand why `anonymousPaths` comes from the `authenticatedPaths` config value

tomeresk (changes requested):

services/src/modules/authentication/strategies/api-key.ts
AleF83: Replaced `enabled` with `disabled`

tomeresk (approved):

PR #319, opened 2021-04-25
@errorHandler directive

AleF83:


services/src/modules/directives/error-handler.ts
```diff
@@ -0,0 +1,125 @@
...
+
+    object.resolveObject = async (source, fields, context, info) => {
+      const injectionArgs = { source, args: {}, context, info };
+      let result;
+      try {
+        result = await originalResolve.call(object, source, fields, context, info);
+      } catch (error) {
+        result = handleCatchError(catchError, error, injectionArgs, result);
+      }
+      throwError && handleThrowError(throwError, result, injectionArgs);
```
tomeresk, line 31: `handleThrowError` is a little confusing here (since it doesn't always throw an error), maybe something like `handleThrowErrorCondition`

services/src/modules/directives/error-handler.ts
```diff
@@ -0,0 +1,125 @@
...
+    condition?: string;
+    errorToThrow: string;
+  };
+};
+
+function handleCatchError(
+  catchError: { condition?: string | undefined; returnValue?: string | Record<string, unknown> } | undefined,
+  error: Error,
+  injectionArgs: GraphQLFieldResolverParams<unknown, RequestContext>,
+  result: unknown
```
tomeresk, line 84: The function shouldn't get this parameter (and doesn't use its value), should just be defined as a variable inside the function

docs/directives/error-handler.md
```diff
@@ -0,0 +1,89 @@
...
+  - `returnValue` (object, optional, default: `null`) - object or string that may be evaluated using [Argument injection mechanism](../arguments_injection.md). This value will be resolved if the error will be caught.
+
+- `throwError` (object, optional, default: `undefined`) - This argument has 2 properties and allows to configure what error will be thrown instead of resolver value.
+  - `condition`: (string, optional, default: `true`) - this expression is evaluated using [Argument injection mechanism](../arguments_injection.md). Additionally to the regular `source`, `args`, `context` and `info` variables, `result` variable is available. Its value is the resolver value or `catchError` section value. See above.
+  - `errorToThrow`: (string, optional, default: empty string). String that may be evaluated using [Argument injection mechanism](../arguments_injection.md). If the evaluation result is `typeof Error` it will be thrown. If the expression cannot be evaluated or is evaluated to string `new Error(result)` will be thrown.
+
+## Execution order
+
+![Flow diagram](../images/error-handler-directive-flow-diagram.png 'Flow Diagram')
```
tomeresk, line 40: At the bottom of the diagram in the `Throw Error` section, `Evaluate <catchError> errorToThrow` should be `Evaluate <throwError> errorToThrow`

tomeresk (changes requested):

tomeresk (approved):

PR #322, opened 2021-04-29
Extend playground config support

bahrambagher:
This change should allow for a more extensive configurable playground. e.g.                 
```
GRAPHQL_PLAYGROUND: | 
                    {
                        "settings": {
                            "request.credentials": "same-origin"
                        }
                     }
```
                     
It would be a breaking change to whoever is adding false to the playground through this syntax 
`GRAPHQL_PLAYGROUND: '"false"'`, which would evaluate to truthy after the change.

To provide a false value it would need to be like this `GRAPHQL_PLAYGROUND: "false"`(single or double quoted)

### DIFF
--- a/services/src/modules/config.ts
+++ b/services/src/modules/config.ts
@@ -25,7 +25,7 @@ export const resourceUpdateInterval = envVarExt.get('RESOURCE_UPDATE_INTERVAL').
 
 // GraphQL configuration
 export const enableGraphQLTracing = envVarExt.get('GRAPHQL_TRACING').default('true').asBool();
-export const enableGraphQLPlayground = envVarExt.get('GRAPHQL_PLAYGROUND').default('true').asBoolStrict();
+export const enableGraphQLPlayground = envVarExt.get('GRAPHQL_PLAYGROUND').default('true').asJsonObject();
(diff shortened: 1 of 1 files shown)


services/src/modules/config.ts
```diff
@@ -25,7 +25,7 @@ export const resourceUpdateInterval = envVarExt.get('RESOURCE_UPDATE_INTERVAL').
 
 // GraphQL configuration
 export const enableGraphQLTracing = envVarExt.get('GRAPHQL_TRACING').default('true').asBool();
-export const enableGraphQLPlayground = envVarExt.get('GRAPHQL_PLAYGROUND').default('true').asBoolStrict();
+export const enableGraphQLPlayground = envVarExt.get('GRAPHQL_PLAYGROUND').default('true').asJsonObject();
```
AleF83, line 28:
```suggestion
export const enableGraphQLPlayground = envVarExt.get('GRAPHQL_PLAYGROUND').default('true').asJsonObject() as PlaygroundConfig;
```
where
```typescript
import { PlaygroundConfig } from 'apollo-server-core';

AleF83 (changes requested):

AleF83 (approved):

PR #323, opened 2021-05-01
Use UnauthorizedByPolicy error in @policies directive

AleF83:


tomeresk (approved):

thewizarodofoz (approved):

services/src/modules/directives/policy/policy.ts
```diff
@@ -15,11 +15,17 @@ export class PolicyDirective extends SchemaDirectiveVisitor {
...
+      if (!context.ignorePolicies && !policy.postValidate) {
         await context.policyExecutor.validatePolicy(policy, source, {}, context, info);
       }
 
-      return originalResolveObject ? await originalResolveObject(source, fields, context, info) : source;
+      const result = originalResolveObject ? await originalResolveObject(source, fields, context, info) : source;
+
+      if (!context.ignorePolicies && policy.postValidate) {
+        await context.policyExecutor.validatePolicy(policy, source, {}, context, info, result);
+      }
```
tomeresk, line 26: Random thought, but since `ignorePolicies` is on the context and we pass the context to `validatePolicy`, maybe we should stop checking `if (!context.ignorePolicies)` here and just call `validatePolicy` anyway, having `validatePolicy` itself not do anything if `ignorePolicies` is set

It will save checking this in quite a few places

tomeresk (changes requested):

docs/authorization.md
```diff
@@ -158,6 +158,41 @@ In the example above, the `roles` argument value is received by using the Argume
 
 > Note: See [Arguments Injection](./arguments_injection.md) for more information about configuration of policy arguments.
 
+## Post validate
```
tomeresk, line 161:
```suggestion
## Post Resolve
```

tomeresk (approved):

PR #326, opened 2021-05-18
fix serialization of rest query key value when object

yoavkarako:


AleF83:
Query param should be string. The only exception is array of strings. I don't think we need to allow/handle other cases.

@tomeresk , WDYT?

services/tests/integration/gateway/rest-directive/directive.spec.ts
```diff
@@ -56,6 +56,30 @@ const testCases: [string, TestCase][] = [
...
+      mock: () =>
+        nock(upstreamHost)
+          .get('/hello')
+          .query({ name: 'miriam' })
+          .reply(200, url => `${new URL(url, upstreamHost).searchParams.get('name')}!`),
+      schema: gql`
+        type Query {
+          hello(name: String!): String! @rest(url: "${upstreamHost}/hello" query: [
+            { key: "name", value: "{args.name}"},
+            { key: "nameObject", value: "{{name: args.name}}"},
```
AleF83, line 71: IMO it should be as following:
```typescript
{ key: "nameObject", value: "{ JSON.stringify({name: args.name}) }"},
```
So we don't need special handling for objects

yoavkarako: I can't imagine a scenario where you wouldn't want `JSON.stringify` an object. The current result will be `[object Object]`.
Might as well make `JSON.stringify` the default.

PR #334, opened 2021-06-19
Fix: Base policy on field w/o directives

AleF83:


services/src/modules/apollo-server-plugins/base-policy.ts
```diff
@@ -37,6 +36,11 @@ export function createBasicPolicyPlugin(): ApolloServerPlugin {
             }
             return policyDefinition.shouldOverrideBasePolicy ?? false;
           });
+          console.log('====== shouldOverrideBasePolicy ====', {
```
thewizarodofoz, line 39: remove?

thewizarodofoz (approved):

PR #337, opened 2021-07-29
introspection query policy

thewizarodofoz:
Currently stitch exposes `GRAPHQL_INTROSPECTION` with a default of `true`.
Enabling the introspection query in production is a security vulnerability in some use cases, so we want to control access to the introspection query using a policy instead.
If the introspection query is not provided, it will be allowed by default (assuming `GRAPHQL_INTROSPECTION` is true).

### DIFF
--- a/services/src/modules/apollo-server-plugins/introspection-query-policy.ts
+++ b/services/src/modules/apollo-server-plugins/introspection-query-policy.ts
@@ -1,22 +1,26 @@
-import { GraphQLResolveInfo, SchemaMetaFieldDef, TypeMetaFieldDef } from 'graphql';
+import { GraphQLField, SchemaMetaFieldDef, TypeMetaFieldDef } from 'graphql';
+import { ApolloServerPlugin } from 'apollo-server-plugin-base';
 import logger from '../logger';
 import { RequestContext } from '../context';
 import { ResourceGroup } from '../resource-repository';
 import { ActiveDirectoryAuth } from '../upstreams/authentication';
 import { PolicyExecutor } from '../directives/policy';
 
-const INTROSPECTION_QUERY_FIELD_NAMES = new Set([SchemaMetaFieldDef.name, TypeMetaFieldDef.name]);
-
-function isIntrospectionRequest(info: GraphQLResolveInfo) {
-  return INTROSPECTION_QUERY_FIELD_NAMES.has(info.fieldName);
+export function createIntrospectionQueryPolicyPlugin(): ApolloServerPlugin {
+  return {
+    serverWillStart() {
+      wrapWithPolicyValidation(SchemaMetaFieldDef);
+      wrapWithPolicyValidation(TypeMetaFieldDef);
+    },
+  };
 }
 
-export function createIntrospectionQueryPolicyPlugin(): void {
-  const originalSchemaMetaFieldResolver = SchemaMetaFieldDef.resolve;
+function wrapWithPolicyValidation(field: GraphQLField<unknown, RequestContext>) {
+  const originalResolver = field.resolve;
 
-  SchemaMetaFieldDef.resolve = function (source, args, context, info) {
+  field.resolve = function (source, args, context, info) {
     function callOriginal() {
-      return originalSchemaMetaFieldResolver!.call(SchemaMetaFieldDef, source, args, context, info);
+      return originalResolver!.call(field, source, args, context, info);
     }
 
     if (!context) {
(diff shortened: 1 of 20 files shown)


services/src/modules/apollo-server-plugins/introspection-query-policy.ts
```diff
@@ -0,0 +1,47 @@
+import { GraphQLResolveInfo, SchemaMetaFieldDef } from 'graphql';
+import { ForbiddenError } from 'apollo-server-core';
+import logger from '../logger';
+import { RequestContext } from '../context';
+import { ResourceGroup } from '../resource-repository';
+import { ActiveDirectoryAuth } from '../upstreams/authentication';
+import { PolicyExecutor } from '../directives/policy';
+
+const INTROSPECTION_QUERY_FIELD_NAMES = new Set(['__schema', '__type']);
```
AleF83, line 9:
```suggestion
const INTROSPECTION_QUERY_FIELD_NAMES = new Set([SchemaMetaFieldDef.name, TypeMetaFieldDef.name]);
```

services/src/modules/apollo-server-plugins/introspection-query-policy.ts
```diff
@@ -0,0 +1,47 @@
...
+
+function isIntrospectionRequest(info: GraphQLResolveInfo) {
+  return INTROSPECTION_QUERY_FIELD_NAMES.has(info.fieldName);
+}
+
+export function createIntrospectionQueryPolicyPlugin(): void {
+  const originalSchemaMetaFieldResolver = SchemaMetaFieldDef.resolve;
```
AleF83, line 16: should be done in global scope to prevent multiple wrapping with policy validation (in tests)
maybe something like this:

```ts
const originalSchemaMetaFieldDefResolve = SchemaMetaFieldDef.resolve;
const originalTypeMetaFieldDefResolve = TypeMetaFieldDef.resolve;

export function createIntrospectionQueryPlugin(): ApolloServerPlugin {
  return {
    serverWillStart() {
      wrapWithPolicyValidation(SchemaMetaFieldDef, originalSchemaMetaFieldDefResolve);
      wrapWithPolicyValidation(TypeMetaFieldDef, originalTypeMetaFieldDefResolve);
    },
  };
}

function wrapWithPolicyValidation(
  field: GraphQLField<unknown, RequestContext>,
  originalResolve?: GraphQLFieldResolver<unknown, RequestContext, Record<string, unknown>>
) {
...
}
```

services/src/modules/apollo-server-plugins/introspection-query-policy.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    if (isIntrospectionRequest(info)) {
+      let policy = introspectionQueryPolicy;
+
+      if (!policy) {
+        logger.info('Introspection query policy not found, using base policy');
+        policy = basePolicy;
+      }
+
+      if (!policy) {
```
AleF83, line 30: We cannot force introspection query authorization. It's open source.

services/src/modules/apollo-server-plugins/introspection-query-policy.ts
```diff
@@ -0,0 +1,47 @@
...
+
+function isIntrospectionRequest(info: GraphQLResolveInfo) {
+  return INTROSPECTION_QUERY_FIELD_NAMES.has(info.fieldName);
+}
+
+export function createIntrospectionQueryPolicyPlugin(): void {
+  const originalSchemaMetaFieldResolver = SchemaMetaFieldDef.resolve;
+
+  SchemaMetaFieldDef.resolve = function (source, args, context, info) {
+    const { resourceGroup, policyExecutor } = (context as unknown) as RequestContext;
```
AleF83, line 19: Apollo server performs introspection query request w/o sending context [here](https://github.com/apollographql/apollo-server/blob/94862a48ff9430d3238f9b456c9a99159fd0aa69/packages/apollo-server-core/src/utils/schemaHash.ts#L12). So the context will be undefined and destructuring will fail. 
So in the case the context is null we need to skip policy check.
I think it's the cause of the infinitive schema update issue you've experienced.

AleF83 (changes requested):

services/src/modules/apollo-server-plugins/introspection-query-policy.ts
```diff
@@ -0,0 +1,57 @@
...
+      return callOriginal();
+    }
+
+    const { resourceGroup, policyExecutor } = context as unknown as RequestContext;
+    const { introspectionQueryPolicy, basePolicy } = resourceGroup;
+
+    let policy = introspectionQueryPolicy;
+
+    if (!policy) {
+      logger.debug('Introspection query policy not found, using base policy');
```
AleF83, line 37:
```suggestion
      logger.trace('Introspection query policy not found, using base policy');
```

services/src/modules/apollo-server-plugins/introspection-query-policy.ts
```diff
@@ -0,0 +1,57 @@
...
+      policy = basePolicy;
+    }
+
+    if (!policy) {
+      logger.debug('Base policy not found, allowing access by default');
```
AleF83, line 42:
```suggestion
      logger.trace('Base policy not found, allowing access by default');
```

services/src/modules/apollo-server-plugins/introspection-query-policy.ts
```diff
@@ -0,0 +1,57 @@
...
+      return originalResolver!.call(field, source, args, context, info);
+    }
+
+    if (!context) {
+      // Apollo is sending introspection queries to generate schema hash, we can't and don't want to block that.
+      return callOriginal();
+    }
+
+    const { resourceGroup, policyExecutor } = context as unknown as RequestContext;
+    const { introspectionQueryPolicy, basePolicy } = resourceGroup;
```
AleF83, line 32:
```suggestion
    const {
      resourceGroup: { introspectionQueryPolicy, basePolicy },
      policyExecutor,
    } = (context as unknown) as RequestContext;
```

AleF83 (commented):
Looks good. 
Graphql Playground performs introspection query every [2 secs by default](https://github.com/graphql/graphql-playground/blob/60f90a910f6df590c6c87cf3f8922a4f61ef454e/packages/graphql-playground-react/src/state/workspace/reducers.ts#L58). So the logging should be on `trace` level.

tomeresk (commented):
Looks good, I agree with the comments about trace logging

AleF83 (approved):

PR #356, opened 2022-07-12
contribution file and eslint alignment

nadavshlezz:
1. Contribution file created - according to soluto policy in order to align soluto repos properly
2. Eslint alignment - discussed with @TheYarin on it - he will go over my PRs on Eslint in order to set up a baseline alignment as much as I can
  *Added rules
  * set default formatter prettier
  * new extension recommendation
  * changedprettierr config file

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,5 +1,9 @@
 {
   "root": true,
+  "env": {
+    "browser": true,
+    "es2021": true
+  },
   "parser": "@typescript-eslint/parser",
   "parserOptions": {
     "ecmaVersion": 2019,
@@ -22,7 +26,8 @@
     "plugin:unicorn/recommended",
     "plugin:prettier/recommended",
     "plugin:promise/recommended",
-    "plugin:jest/style"
+    "plugin:jest/style",
+    "plugin:eslint-comments/recommended"
   ],
   "rules": {
     "no-process-exit": "off",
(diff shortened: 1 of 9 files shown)


AleF83 (dismissed):

.eslintrc
```diff
@@ -1,5 +1,9 @@
 {
   "root": true,
+  "env": {
+    "browser": true,
```
tomeresk, line 4: Why would we set `browser` true for a nodejs/backend project?

nadavshlezz: yes - my mistake :) changed it

tomeresk (dismissed):

.vscode/settings.json
```diff
@@ -1,3 +1,4 @@
 {
-    "debug.node.autoAttach": "on"
+  "debug.node.autoAttach": "on",
+  "editor.defaultFormatter": "esbenp.prettier-vscode"
 }
```
TheYarin, line 4: I would also add the "use `.gitignore`" setting.

.eslintrc
```diff
@@ -1,17 +1,15 @@
 {
   "root": true,
+  "env": {
+    "node": true,
+    "es2021": true
```
TheYarin, line 5: why not 2022?

TheYarin (approved):

.eslintrc
nadavshlezz: image

nadavshlezz: according to eslint repo - we cant change it right now - it does not work or will effect - me and yarin have decided to leave it like that so the ecmaVersion should be coordinated with the node/typescript version

AleF83 (approved):

